PR #23, opened 2022-05-21
Add the heap data-structure

QGarchery:
This implements a heap as the data structure for the matching engine. The implementation uses a "basic" heap in the sense that its elements are simply stored in an array with the following rules for the indexes:

- they start from index `1` and end at the length of the array (included)
- index `0` is used for checking if elements are not in the heap
- from index `i`, go to the left child with `2 * i`
- from index `i`, go to the right child with `2 * i + 1`
- from index `i`, go to the parent with `i / 2`

There is an address as satellite data `id`, that can be used, for example, to retrieve whose account has the biggest amount. There is also a mapping `indexes` from addresses to indexes in the array, allowing us to access the account associated to a given address.

The only non-view, non-private function is the function `update`. According to the parameters it receives, it calls one of the private functions `insert`, `decrease`, `increase` or `remove`.

### DIFF
--- a/contracts/Heap.sol
+++ b/contracts/Heap.sol
@@ -123,8 +123,10 @@ library BasicHeap {
             maxIndex = _index;
             maxValue = _heap.accounts[_index - 1].value;
 
-            if (leftIndex <= accountsLength && _heap.accounts[leftIndex - 1].value > maxValue)
+            if (leftIndex <= accountsLength && _heap.accounts[leftIndex - 1].value > maxValue) {
                 maxIndex = leftIndex;
+                maxValue = _heap.accounts[leftIndex - 1].value;
+            }
 
             if (rightIndex <= accountsLength && _heap.accounts[rightIndex - 1].value > maxValue)
                 maxIndex = rightIndex;
--- a/contracts/Heap.sol
+++ b/contracts/Heap.sol
@@ -3,13 +3,13 @@ pragma solidity ^0.8.0;
 
 library BasicHeap {
     struct Account {
-        address id;
-        uint256 value;
+        address id; // The address of the account.
+        uint256 value; // The value of the account.
     }
 
     struct Heap {
-        Account[] accounts;
-        mapping(address => uint256) indexes;
+        Account[] accounts; // All the accounts.
+        mapping(address => uint256) ranks; // A mapping from an address to a rank in accounts. Beware: ranks are shifted by one compared to indexes, so the first rank is 1 and not 0.
     }
 
     /// ERRORS ///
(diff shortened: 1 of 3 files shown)


contracts/Heap.sol
```diff
@@ -0,0 +1,236 @@
...
+        uint256 index = _heap.indexes[_id];
+        if (index < _heap.accounts.length) return _heap.accounts[index].id;
+        else return address(0);
+    }
+
+    /// PRIVATE ///
+
+    /// @notice Sets `_index` in the `_heap` to be `_account`.
+    /// @dev The heap may lose its invariant about the order of the values stored.
+    /// @dev Only call this function with an index in the bounds of the array.
```
MerlinEgalite, line 79: still don't this is necessary as it would revert

contracts/Heap.sol
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    /// @notice Moves an account up the heap until its value is smaller than the one of its parent.
+    /// @dev This functions restores the invariant about the order of the values stored when the account at `_index` is the only one with value greater than what it should be.
+    /// @param _heap The heap to modify.
+    /// @param _index The index of the account to move.
+    function shiftUp(Heap storage _heap, uint256 _index) private {
+        Account memory initAccount = _heap.accounts[_index - 1];
+        uint256 initValue = initAccount.value;
+        while (_index != 1 && initValue > _heap.accounts[_index / 2 - 1].value) {
```
MerlinEgalite, line 116: did you try the bit shift for all mul and iv by 2?

contracts/Heap.sol
```diff
@@ -0,0 +1,236 @@
...
+    /// @param _index The index of the account to move.
+    function shiftDown(Heap storage _heap, uint256 _index) private {
+        uint256 size = _heap.accounts.length;
+        Account memory initAccount = _heap.accounts[_index - 1];
+        uint256 childIndex = _index * 2;
+        Account memory childAccount;
+
+        while (childIndex <= size) {
+            if (
+                childIndex + 1 <= size &&
```
MerlinEgalite, line 135: could still be otpimized

MerlinEgalite (approved):
Small comments bur overall LGTM

QGarchery:
After thinking about it, it should be made more general. For example the `remove`, `insert`, `increase` and `decrease` should be made internal.
I'll leave this PR like this for the moment

pakim249CAL (approved):

QGarchery:
We should also provide a way to abstract the satellite data in the structure

QGarchery:
We can also implement:
- #38 
- #36

Rubilmax (approved):

PR #25, opened 2022-05-29
Heap plus unordered data structure

QGarchery:
See #23 for an explanation of the heap indexes.

There is now a new parameter `size` in the heap data-structure. The heap consists of the elements of index 1 through `size`.

The size grows at insertion but should never be greater than `maxSortedUsers`. To that end, the size is divided by 2 every time it hits that limit, removing only the leafs.

### DIFF
--- a/contracts/HeapOrdering.sol
+++ b/contracts/HeapOrdering.sol
@@ -10,7 +10,7 @@ library HeapOrdering {
     struct HeapArray {
         Account[] accounts; // All the accounts.
         uint256 size; // The size of the heap portion of the structure, should be less than accounts length, the rest is an unordered array.
-        mapping(address => uint256) indexes; // A mapping from an address to an index in accounts.
+        mapping(address => uint256) ranks; // A mapping from an address to an rank in accounts.
     }
 
     /// ERRORS ///
@@ -32,9 +32,9 @@ library HeapOrdering {
     /// @param _id The address of the account.
     /// @return The value of the account.
     function getValueOf(HeapArray storage _heap, address _id) internal view returns (uint256) {
-        uint256 index = _heap.indexes[_id];
-        if (index == 0) return 0;
-        else return _heap.accounts[index - 1].value;
+        uint256 rank = _heap.ranks[_id];
+        if (rank == 0) return 0;
+        else return _heap.accounts[rank - 1].value;
     }
 
     /// @notice Returns the address at the head of the `_heap`.
@@ -59,8 +59,8 @@ library HeapOrdering {
     /// @param _id The address of the account.
     /// @return The address of the previous account.
     function getPrev(HeapArray storage _heap, address _id) internal view returns (address) {
-        uint256 index = _heap.indexes[_id];
-        if (index > 1) return _heap.accounts[index - 2].id;
+        uint256 rank = _heap.ranks[_id];
+        if (rank > 1) return _heap.accounts[rank - 2].id;
         else return address(0);
     }
 
@@ -70,8 +70,8 @@ library HeapOrdering {
     /// @param _id The address of the account.
     /// @return The address of the next account.
     function getNext(HeapArray storage _heap, address _id) internal view returns (address) {
-        uint256 index = _heap.indexes[_id];
-        if (index < _heap.accounts.length) return _heap.accounts[index].id;
+        uint256 rank = _heap.ranks[_id];
+        if (rank < _heap.accounts.length) return _heap.accounts[rank].id;
         else return address(0);
     }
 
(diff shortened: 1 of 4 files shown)


contracts/HeapOrdering.sol
```diff
@@ -0,0 +1,285 @@
...
+    function shiftDown(HeapArray storage _heap, uint256 _rank) private {
+        uint256 size = _heap.size;
+        Account memory initAccount = _heap.accounts[_rank - 1];
+        uint256 childRank = _rank * 2;
+        Account memory childAccount;
+
+        while (childRank <= size) {
+            if (
+                // Compute the rank of the child with biggest value.
+                childRank + 1 <= size &&
```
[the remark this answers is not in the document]
MerlinEgalite: I agree

contracts/HeapOrdering.sol
```diff
@@ -0,0 +1,318 @@
...
+
+library HeapOrdering {
+    struct Account {
+        address id; // The address of the account.
+        uint256 value; // The value of the account.
+    }
+
+    struct HeapArray {
+        Account[] accounts; // All the accounts.
+        uint256 size; // The size of the heap portion of the structure, should be less than accounts length, the rest is an unordered array.
```
pakim249CAL, line 12: So this would kind of be a bit of a hacky non-standard thing, but I believe that gas savings are possible by converting size to a uint96 and packing it with id into the accounts struct at accounts[0]. This way, size can share a storage slot with address. Maybe worth investigating the possibility?

contracts/HeapOrdering.sol
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /// @notice Moves an account up the heap until its value is smaller than the one of its parent.
+    /// @dev This functions restores the invariant about the order of the values stored when the account at `_rank` is the only one with value greater than what it should be.
+    /// @param _heap The heap to modify.
+    /// @param _rank The rank of the account to move.
+    function shiftUp(HeapArray storage _heap, uint256 _rank) private {
+        Account memory initialAccount = getAccount(_heap, _rank);
+        uint256 initialValue = initialAccount.value;
+        while (_rank > 1 && initialValue > getAccount(_heap, _rank >> 1).value) {
```
MerlinEgalite, line 128: ah coole shift!

contracts/HeapOrdering.sol
```diff
@@ -0,0 +1,318 @@
...
+        return _size;
+    }
+
+    /// @notice Returns the account of rank `_rank`.
+    /// @dev The first rank is 1 and the last one is length of the array.
+    /// @dev Only call this function with positive numbers.
+    /// @param _heap The heap to search in.
+    /// @param _rank The rank of the account.
+    /// @return The account of rank `_rank`.
+    function getAccount(HeapArray storage _heap, uint256 _rank)
```
MerlinEgalite, line 67: still don't find this function necessary. Moreover it differs from the other implementation which make comparison worse. `Heap` and `HeapOrdering` should be the most similar as possible without comprising on gas cost IMO

contracts/HeapOrdering.sol
```diff
@@ -0,0 +1,318 @@
...
+        return _heap.accounts[_rank - 1];
+    }
+
+    /// @notice Sets the value at `_rank` in the `_heap` to be `_newValue`.
+    /// @dev The heap may lose its invariant about the order of the values stored.
+    /// @dev Only call this function with a rank within array's bounds.
+    /// @param _heap The heap to modify.
+    /// @param _rank The rank of the account in the heap to be set.
+    /// @param _newValue The new value to set the `_rank` to.
+    function setAccountValue(
```
MerlinEgalite, line 81: same for this function

contracts/HeapOrdering.sol
```diff
@@ -0,0 +1,318 @@
...
+            else shiftUp(_heap, rank);
+        }
+    }
+
+    /// GETTERS ///
+
+    /// @notice Returns the number of users in the `_heap`.
+    /// @param _heap The heap parameter.
+    /// @return The length of the heap.
+    function length(HeapArray storage _heap) internal view returns (uint256) {
```
MerlinEgalite, line 267: why difference where between private and internal?

contracts/HeapOrdering.sol
```diff
@@ -0,0 +1,318 @@
...
+    function decrease(
+        HeapArray storage _heap,
+        address _id,
+        uint256 _newValue
+    ) private {
+        uint256 rank = _heap.ranks[_id];
+        setAccountValue(_heap, rank, _newValue);
+
+        // We only need to restore the invariant if the account is a node in the heap
+        if (rank <= _heap.size >> 1) shiftDown(_heap, rank);
```
MerlinEgalite, line 207: cool

MerlinEgalite (changes requested):
some questions and comments

contracts/HeapOrdering.sol
QGarchery: It's exactly the same in terms of gas consumption (I tested it to be sure), but the private functions are not designed to be used directly by other contracts using this data-structure. For example, the `remove` function assumes the given address corresponds to an account in the heap; it does not check that this is the case with an assert. This is sound because we only use the remove function in our context where we know that those assumptions hold

contracts/HeapOrdering.sol
QGarchery: See #30

contracts/HeapOrdering.sol
pakim249CAL: I suggested this because it improves the readability of the code. Seeing account[rank-1] everywhere is kind of a nightmare for auditing purposes imo, and it doesn't add much gas cost.

GabrielAstieres (approved):
Well played Quentin :)

test-foundry/TestRandomHeapOrdering.t.sol
```diff
@@ -0,0 +1,119 @@
...
+
+    function clearHeap() public {
+        // the next operations will be performed on a new empty heap
+        heaps.push();
+    }
+
+    function updateHeap(
+        address _id,
+        uint256 _newValue,
+        uint256 _max_sorted_users
```
MerlinEgalite, line 46:
```suggestion
        uint256 _maxSortedUsers
```

test-foundry/TestRandomHeapOrdering.t.sol
```diff
@@ -0,0 +1,119 @@
...
+}
+
+contract Helper is Random {
+    HeapOrdering.HeapArray[] internal heaps;
+
+    function setUp() public {
+        heaps.push();
+    }
+
+    function getCurrentHeap() private view returns (HeapOrdering.HeapArray storage) {
```
MerlinEgalite, line 34: not sure about the naming of this function

test-foundry/TestRandomHeapOrdering.t.sol
```diff
@@ -0,0 +1,119 @@
...
+                        user.value >= childUser.value,
+                        "Error: The heap is not correctly ordered."
+                    );
+                }
+            }
+        }
+    }
+}
+
+contract TestHeapMEP is DSTest {
```
MerlinEgalite, line 85: to rename?

test-foundry/TestRandomHeapOrdering.t.sol
```diff
@@ -0,0 +1,119 @@
...
+    Helper public helper = new Helper();
+
+    function setUp() public {
+        helper.setUp();
+    }
+
+    function _stressTest(uint256 n) public {
```
MerlinEgalite, line 92:
```suggestion
    function stressTest(uint256 n) public {
```

test-foundry/TestRandomHeapOrdering.t.sol
```diff
@@ -0,0 +1,119 @@
...
+    function getSize() private view returns (uint256) {
+        HeapOrdering.HeapArray storage heap = getCurrentHeap();
+        return heap.size;
+    }
+
+    function checkOrdering() public view {
+        // Check that for the parent value (at rank i) is always greater than the ones of his children
+        // (at rank 2i, 2i + 1), for ranks <= head size
+        HeapOrdering.HeapArray storage heap = getCurrentHeap();
+        for (uint256 rank = 1; rank < getSize(); rank++) {
```
MerlinEgalite, line 68: getSize outside the for loop to avoid useless computation

test-foundry/TestRandomHeapOrdering.t.sol
```diff
@@ -0,0 +1,119 @@
...
+
+    function setUp() public {
+        helper.setUp();
+    }
+
+    function _stressTest(uint256 n) public {
+        uint256 N_USERS = n;
+        uint256 RANGE_VALUES = 2 * n;
+        uint256 N_ITERS = 3 * n;
+        uint256 max_sorted_users = helper.randomUint256(N_USERS);
```
MerlinEgalite, line 96: camelCase 🐫

MerlinEgalite (approved):
Small suggestions but overall looks good!

PR #26, opened 2022-05-31
Add uncheckeds to rank computation

Rubilmax:
I don't think it's worth it, but the work needs to be seen though
It can be seen "risky" but if my maths are correct it's 100% secure lol


QGarchery:
I have stolen most of the work of this PR, sorry for that. The rest is the unchecked computations, see #26

PR #31, opened 2022-06-05
Pass size in memory

pakim249CAL:
before:
image

after:
image


QGarchery:
I think it"s a bit too intrusive to make the cut but it's still good to have in mind :)

PR #44, opened 2022-07-05
Report fixes

MathisGD:


contracts/HeapOrdering.sol
```diff
@@ -13,6 +13,10 @@ library HeapOrdering {
         mapping(address => uint256) ranks; // A mapping from an address to a rank in accounts. Beware: ranks are shifted by one compared to indexes, so the first rank is 1 and not 0.
     }
 
+    /// CONSTANTS ///
+
+    uint256 public constant ROOT = 1;
```
QGarchery, line 18: I don't think we want to make it public: it is an implementation detail

QGarchery (approved):
LGTM

contracts/HeapOrdering.sol
MathisGD: Ok, changing it

MerlinEgalite: we don't really care no? We can keep it public no? We're building ob public blockchains 🤓

MerlinEgalite (approved):
LGTM!

contracts/HeapOrdering.sol
QGarchery: Yes sure, it's a really small change, but it looks a bit ugly if that is the only thing public in this contract, don't you think ?

MathisGD: Yes I agree

QGarchery (approved):

Issue #37, opened 2022-07-05
HOG-06C: Removal Swap Optimization


MerlinEgalite:
## HOG-06C: Removal Swap Optimization

| Type | Severity | Location |
| ---- | -------- | -------- |
| [Gas Optimization](/appendix/finding-types#gas-optimization) | <img class="o-severity o-informational" src="https://omniscia.io/report-assets/informational.png" /> | [HeapOrdering.sol:L250](https://github.com/morpho-labs/data-structures/blob/6b3f07abe798c6e4378f80bee76db6b8f3a44ed1/contracts/HeapOrdering.sol#L250) |

### Description:

The `swap` operation performed during an removal operation can be optimized by performing it conditionally only when `rank != accountsLength` as otherwise the heap will replace an element with itself.

### Example:

```solidity
/// @notice Removes an account in the `_heap`.
/// @dev Only call when this function `_id` is in the `_heap` with value `_removedValue`.
/// @param _heap The heap to modify.
/// @param _id The address of the account to remove.
/// @param _removedValue The value of the account to remove.
function remove(
    HeapArray storage _heap,
    address _id,
    uint256 _removedValue
) private {
    uint256 rank = _heap.ranks[_id];
    uint256 accountsLength = _heap.accounts.length;

    // Swap the last account and the account to remove, then pop it.
    swap(_heap, rank, accountsLength);
    if (_heap.size == accountsLength) _heap.size--;
    _heap.accounts.pop();
    delete _heap.ranks[_id];

    // If the swapped account is in the heap, restore the invariant: its value can be smaller or larger than the removed value.
    if (rank <= _heap.size) {
        if (_removedValue > getAccount(_heap, rank).value) shiftDown(_heap, rank);
        else shiftUp(_heap, rank);
    }
}
```

### Recommendation:

We advise this conditional to be introduced optimizing the best-case gas cost of the function with minimal impact on the worst-case scenario.

Issue #34, opened 2022-07-05
Increase Swap Optimization


MerlinEgalite:
## HOG-02C: Increase Swap Optimization

| Type | Severity | Location |
| ---- | -------- | -------- |
| [Gas Optimization](/appendix/finding-types#gas-optimization) | <img class="o-severity o-informational" src="https://omniscia.io/report-assets/informational.png" /> | [HeapOrdering.sol:L230](https://github.com/morpho-labs/data-structures/blob/6b3f07abe798c6e4378f80bee76db6b8f3a44ed1/contracts/HeapOrdering.sol#L230) |

### Description:

The `swap` operation performed during an increase operation can be optimized by performing it conditionally only when `nextSize != rank` as otherwise the heap will replace an element with itself.

### Example:

```solidity
/// @notice Increases the amount of an account in the `_heap`.
/// @dev Only call this function when `_id` is in the `_heap` with a smaller value than `_newValue`.
/// @param _heap The heap to modify.
/// @param _id The address of the account to increase the amount.
/// @param _newValue The new value of the account.
/// @param _maxSortedUsers The maximum size of the heap.
function increase(
    HeapArray storage _heap,
    address _id,
    uint256 _newValue,
    uint256 _maxSortedUsers
) private {
    uint256 rank = _heap.ranks[_id];
    Account memory account = getAccount(_heap, rank);
    account.value = _newValue;
    setAccountValue(_heap, rank, _newValue);
    uint256 nextSize = _heap.size + 1;

    if (rank < nextSize) shiftUp(_heap, rank);
    else {
        swap(_heap, nextSize, rank);
        shiftUp(_heap, nextSize);
        _heap.size = computeSize(nextSize, _maxSortedUsers);
    }
}
```

### Recommendation:

We advise this conditional to be introduced optimizing the best-case gas cost of the function with minimal impact on the worst-case scenario.

PR #45, opened 2022-07-06
Optimize swap operation for the same ranks

### LINKED ISSUE #32: HOG-03C: Insertion Swap Optimization
## HOG-03C: Insertion Swap Optimization

| Type | Severity | Location |
| ---- | -------- | -------- |
| [Gas Optimization](/appendix/finding-types#gas-optimization) |  | [HeapOrdering.sol:L188](https://github.com/morpho-labs/data-structures/blob/6b3f07abe798c6e4378f80bee76db6b8f3a44ed1/contracts/HeapOrdering.sol#L188) |

### Description:

The `swap` operation performed during an insertion operation can be optimized by performing it conditionally only when `newSize != accountsLength` as otherwise the heap will replace an element with itself.

### Example:

```solidity
/// @notice Inserts an account in the `_heap`.
/// @dev Only call this function when `_id` is not in the `_heap`.
/// @dev Reverts with AddressIsZero if `_value` is 0.
/// @param _id The address of the account to insert.
/// @param _value The value of the account to insert.
/// @param _maxSortedUsers The maximum size of the heap.
function insert(
    HeapArray storage _heap,
    address _id,
    uint256 _value,
    uint256 _maxSortedUsers
) private {
    // `_heap` cannot contain the 0 address.

    _heap.accounts.push(Account(_id, _value));
    _heap.ranks[_id] = accountsLength;

    // Move the account at the end of the heap and restore the invariant.
    uint256 newSize = _heap.size + 1;
    swap(_heap, newSize, accountsLength);
    shiftUp(_heap, newSize);
    _heap.size = computeSize(newSize, _maxSortedUsers);
}
```

### Recommendation:

We advise this conditional to be introduced optimizing the best-case gas cost of the function with minimal impact on the worst-case scenario.

QGarchery:
Fixes
- #32 
- #34 
- #37

Before the changes (only significant function is `update`):
before-swap-same
After the changes:
after-swap-same

### DIFF
--- a/contracts/HeapOrdering.sol
+++ b/contracts/HeapOrdering.sol
@@ -112,6 +112,7 @@ library HeapOrdering {
         uint256 _rank1,
         uint256 _rank2
     ) private {
+        if (_rank1 == _rank2) return;
         Account memory accountOldRank1 = getAccount(_heap, _rank1);
         Account memory accountOldRank2 = getAccount(_heap, _rank2);
         setAccount(_heap, _rank1, accountOldRank2);
--- a/test-foundry/TestHeapOrdering.t.sol
+++ b/test-foundry/TestHeapOrdering.t.sol
@@ -503,4 +503,38 @@ contract TestHeapOrdering is DSTest {
 
         assertEq(heap.accounts[1].value, 25);
     }
+
+    function testInsertNoSwap() public {
+        update(accounts[0], 0, 40);
+        update(accounts[1], 0, 30);
+        update(accounts[2], 0, 20);
+
+        // Insert does a swap with the same ranks.
+        update(accounts[3], 0, 10);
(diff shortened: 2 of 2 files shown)


MathisGD (commented):
If you think about it, there is a very low chance that this happens. And the bigger the heap is, the lower the chance is that you should not be swapped. It also seems that there is no realistic scenario (for example doing multiple operations one after another) that makes this scenario happen.

This is why, as we can see with test, this is not an optimization on average (because of the cost of the if ==). We might not implement this.

QGarchery:
Changing tests to only the one with 10000 possible values:
before

after the changes of this PR:
after-swap-same

MathisGD (approved):
LGTM this time

Issue #33, opened 2022-07-05
Downward Traversal Optimization


MerlinEgalite:
## HOG-01C: Downward Traversal Optimization

| Type | Severity | Location |
| ---- | -------- | -------- |
| [Gas Optimization](/appendix/finding-types#gas-optimization) | <img class="o-severity o-informational" src="https://omniscia.io/report-assets/informational.png" /> | [HeapOrdering.sol:L151](https://github.com/morpho-labs/data-structures/blob/6b3f07abe798c6e4378f80bee76db6b8f3a44ed1/contracts/HeapOrdering.sol#L151), [L154](https://github.com/morpho-labs/data-structures/blob/6b3f07abe798c6e4378f80bee76db6b8f3a44ed1/contracts/HeapOrdering.sol#L154) |

### Description:

The `shiftDown` downward traversal mechanism of the heap structure is sub-optimal as it computes the result of `getAccount` operations duplicate times in all cases.

### Example:

```solidity
while (childRank <= size) {
    // Compute the rank of the child with largest value.
    if (
        childRank < size &&
        getAccount(_heap, childRank + 1).value > getAccount(_heap, childRank).value
    ) childRank++;

    childAccount = getAccount(_heap, childRank);

    if (childAccount.value > initialValue) {
        setAccount(_heap, _rank, childAccount);
        _rank = childRank;
        childRank <<= 1;
    } else break;
}
```

### Recommendation:

We advise the result of the first `getAccount` invocation within the `while` loop to be cached and overwritten in the `if` block by the next child's account lookup if the condition is met, thus ensuring `getAccount` is at most invoked a single time per child.

PR #46, opened 2022-07-06
Downward traversal optimization

QGarchery:
Fixes 
- #33 

before-swap-same

After changes:
after-optimize-get

### DIFF
--- a/contracts/HeapOrdering.sol
+++ b/contracts/HeapOrdering.sol
@@ -152,7 +152,9 @@ library HeapOrdering {
             if (childRank < size) {
                 rightChild = getAccount(_heap, childRank + 1);
                 if (rightChild.value > childToSwap.value) {
-                    childRank++;
+                    unchecked {
+                        ++childRank; // This cannot overflow because childRank < size.
+                    }
                     childToSwap = rightChild;
                 }
             }
--- a/contracts/HeapOrdering.sol
+++ b/contracts/HeapOrdering.sol
@@ -141,17 +141,21 @@ library HeapOrdering {
         Account memory accountToShift = getAccount(_heap, _rank);
         uint256 valueToShift = accountToShift.value;
         Account memory childToSwap;
+        Account memory rightChild;
         uint256 childRank = _rank << 1;
         // At this point, childRank (resp. childRank+1) is the rank of the left (resp. right) child.
 
         while (childRank <= size) {
-            // Compute the rank of the child with largest value.
-            if (
-                childRank < size &&
-                getAccount(_heap, childRank + 1).value > getAccount(_heap, childRank).value
-            ) childRank++;
-
             childToSwap = getAccount(_heap, childRank);
+
+            // Find the child with largest value.
+            if (childRank < size) {
+                rightChild = getAccount(_heap, childRank + 1);
+                if (rightChild.value > childToSwap.value) {
+                    childRank++;
+                    childToSwap = rightChild;
+                }
+            }
 
             if (childToSwap.value > valueToShift) {
                 setAccount(_heap, _rank, childToSwap);


contracts/HeapOrdering.sol
```diff
@@ -141,18 +141,22 @@ library HeapOrdering {
         Account memory initialAccount = getAccount(_heap, _rank);
         uint256 initialValue = initialAccount.value;
         Account memory childAccount;
+        Account memory otherChildAccount;
```
MathisGD, line 144: rightChild ?

contracts/HeapOrdering.sol
```diff
@@ -141,18 +141,22 @@ library HeapOrdering {
         Account memory initialAccount = getAccount(_heap, _rank);
         uint256 initialValue = initialAccount.value;
         Account memory childAccount;
```
MathisGD, line 143: childToSwap ?

contracts/HeapOrdering.sol
```diff
@@ -141,18 +141,22 @@ library HeapOrdering {
         Account memory initialAccount = getAccount(_heap, _rank);
         uint256 initialValue = initialAccount.value;
```
MathisGD, line 142: not sure why it is called "initial".

MathisGD (commented):
Again, this in practice does not seem to be a real optimization (on our test base), but I don't really understand why. We remove one warm SLOAD but add one if, and one account in memory.

If it stays like that, again I don't see why we would do that.

contracts/HeapOrdering.sol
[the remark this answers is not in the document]
QGarchery: It's used only one time in the code but we look at this value at each iteration. So it should save the lookup of the `value` field everytime

QGarchery: It's the account we start with, but I agree that it is not the best naming

contracts/HeapOrdering.sol
QGarchery: I'm ok with it. Not changing it in this PR though because we are not sure we want to merge it

QGarchery: See #50

contracts/HeapOrdering.sol
QGarchery: See #50

contracts/HeapOrdering.sol
```diff
@@ -141,18 +141,22 @@ library HeapOrdering {
         Account memory accountToShift = getAccount(_heap, _rank);
         uint256 valueToShift = accountToShift.value;
         Account memory childToSwap;
+        Account memory rightChild;
```
MerlinEgalite, line 144: did you try internalizing those avariables inside the loop? i know it sounds a bit weird but sometimes it saves some gas

contracts/HeapOrdering.sol
```diff
@@ -141,18 +141,22 @@ library HeapOrdering {
...
         // At this point, childRank (resp. childRank+1) is the rank of the left (resp. right) child.
 
         while (childRank <= size) {
-            // Compute the rank of the child with largest value.
-            if (
-                childRank < size &&
-                getAccount(_heap, childRank + 1).value > getAccount(_heap, childRank).value
-            ) childRank++;
-
             childToSwap = getAccount(_heap, childRank);
```
MerlinEgalite, line 149: I guess the gas increase comes from this assignation which is done even if `(childRank < size)` which was not done before

contracts/HeapOrdering.sol
```diff
@@ -141,18 +141,22 @@ library HeapOrdering {
...
 
+            // Find the child with largest value.
+            if (childRank < size) {
+                rightChild = getAccount(_heap, childRank + 1);
+                if (rightChild.value > childToSwap.value) {
+                    childRank++;
```
MerlinEgalite, line 155:
```suggestion
                    unchecked {
                        ++childRank;
                    }
```

to save gas

MerlinEgalite (changes requested):
changes to try

contracts/HeapOrdering.sol
QGarchery: This variable is multiplied by 2 many times, so I don't think it applies safely here (even if it probably does, because childrank is related to a rank in an array). It is completely safe for variables that are only incremented, because you know you won't ever be able to reach 2^256 by only incrementing

contracts/HeapOrdering.sol
QGarchery: Sorry but I won't do it without a rationale behind it. Otherwise we could spend a lot of time not understanding what we are doing

QGarchery:
Changing tests to only the one with 10000 entries:
before

after the changes of this PR:
after-opti-shiftdown

contracts/HeapOrdering.sol
QGarchery: After talking with @MathisGD, I think this is safe actually because we know that `childRank < size`. So it can't overflow

MathisGD (approved):

contracts/HeapOrdering.sol
MerlinEgalite: the rational is that if we don't enter the loop we don't instantiate something useless in memory

MerlinEgalite: but ok with your choice

contracts/HeapOrdering.sol
```diff
@@ -141,18 +141,24 @@ library HeapOrdering {
...
-            ) childRank++;
-
             childToSwap = getAccount(_heap, childRank);
 
+            // Find the child with largest value.
+            if (childRank < size) {
+                rightChild = getAccount(_heap, childRank + 1);
+                if (rightChild.value > childToSwap.value) {
+                    unchecked {
+                        ++childRank; // This cannot overflow because childRank < size.
```
MerlinEgalite, line 156: I don't understand the comment. It cannot overflow because adding up just one at each step would not even fit inside a human's lifespan right?

MerlinEgalite (changes requested):
Perhaps a comment to clarify

contracts/HeapOrdering.sol
MerlinEgalite: it depends on the frequency obviously but you get the idea

QGarchery: We have `childRank < size` and `size <= type(uint256).max` (because `size` is of type `uint256`). So `childRank < type(uint256).max`, this means that it cannot overflow by adding one

QGarchery: Previous comment about `childRank`:
> This variable is multiplied by 2 many times, so I don't think it applies safely here (even if it probably does, because childrank is related to a rank in an array). It is completely safe for variables that are only incremented, because you know you won't ever be able to reach 2^256 by only incrementing

MerlinEgalite: my bad my mistake was I forgot we double the value at each step!

MerlinEgalite (approved):
LGTM

Issue #38, opened 2022-07-05
HOG-07C: Removal Upward Shift Optimization


MerlinEgalite:
## HOG-07C: Removal Upward Shift Optimization

| Type | Severity | Location |
| ---- | -------- | -------- |
| [Gas Optimization](/appendix/finding-types#gas-optimization) | <img class="o-severity o-informational" src="https://omniscia.io/report-assets/informational.png" /> | [HeapOrdering.sol:L258](https://github.com/morpho-labs/data-structures/blob/6b3f07abe798c6e4378f80bee76db6b8f3a44ed1/contracts/HeapOrdering.sol#L258) |

### Description:

The `shiftUp` operation is performed unconditionally whilst a child and parent's value can indeed match, thus executing an upwards shift inefficiently if the element removed was next-to-last in the tree and its children had an equal value to it.

### Example:

```solidity
// If the swapped account is in the heap, restore the invariant: its value can be smaller or larger than the removed value.
if (rank <= _heap.size) {
    if (_removedValue > getAccount(_heap, rank).value) shiftDown(_heap, rank);
    else shiftUp(_heap, rank);
}
```

### Recommendation:

We advise the `else` statement to be adjusted to an `else if (_removedValue < getAccount(_heap, rank).value)` one, with the `getAccount(_heap, rank).value` value being cached to a local variable within the upper-most `if` clause to avoid the duplicate read gas cost between the `if` and `else if` conditional evaluations.

QGarchery:
> These changes seems unlikely to save gas in our case: the amounts will all be different because of the interests earn. So even if 2 person supply exactly the same amount, if one does so one block after the other, the amounts will be different.

PR #48, opened 2022-07-06
Removal upward shift optimization

QGarchery:
Fixes
- #38 

These changes seems unlikely to save gas in our case: the amounts will all be different because of the interests earn. So even if 2 person supply exactly the same amount, if one does so one block after the other, the amounts will be different.

before-swap-same

After changes:
after-remove-shiftup


MerlinEgalite (approved):
LGTM

QGarchery:
Changing tests to only the one with 10000 possible values:
before

after the changes of this PR:
after-remove-shiftup

Issue #35, opened 2022-07-05
HOG-04C: Potential Significant Optimization


MerlinEgalite:
## HOG-04C: Potential Significant Optimization

| Type | Severity | Location |
| ---- | -------- | -------- |
| [Gas Optimization](/appendix/finding-types#gas-optimization) | <img class="o-severity o-informational" src="https://omniscia.io/report-assets/informational.png" /> | [HeapOrdering.sol:L7](https://github.com/morpho-labs/data-structures/blob/6b3f07abe798c6e4378f80bee76db6b8f3a44ed1/contracts/HeapOrdering.sol#L7) |

### Description:

Depending on the needs of the application that will ultimately utilize the library, the `Account` data structure can be significantly optimized by reducing its `value` member's accuracy from 256-bits to 96-bits, thereby permitting the `address` variable to be tight-packed with the `uint96` into a single 32-byte slot. This will reduce all storage reads and writes for `Account` entries to one `SLOAD` / `SSTORE` operation regardless of whether both members are written to or read from in the same action.

### Example:

```solidity
struct Account {
    address id; // The address of the account.
    uint256 value; // The value of the account.
}
```

### Recommendation:

We advise this to be done so if the 96-bit accuracy is sufficient for the end-user application.

Issue #43, opened 2022-07-05
HOG-02M: Inexistent Association of Former Value


QGarchery:
## HOG-02M: Inexistent Association of Former Value

| Type | Severity | Location |
| ---- | -------- | -------- |
| [Input Sanitization](/appendix/finding-types#input-sanitization) | <img class="o-severity o-minor" src="https://omniscia.io/report-assets/minor.png" /> | [HeapOrdering.sol:L30-L36](https://github.com/morpho-labs/data-structures/blob/6b3f07abe798c6e4378f80bee76db6b8f3a44ed1/contracts/HeapOrdering.sol#L30-L36) |

### Description:

The `_formerValue` argument is meant to indicate the current value of the particular `_id` in the heap structure, however, this is not validated and is instead assumed to be trusted input.

### Impact:

An improper `_formerValue` combined with an improper `_newValue` will cause an incorrect operation to be performed on the heap that does not match the actual data the `_id` entry previously held.

### Example:

```solidity
function update(
    HeapArray storage _heap,
    address _id,
    uint256 _formerValue,
    uint256 _newValue,
    uint256 _maxSortedUsers
) internal {
    uint256 size = _heap.size;
    uint256 newSize = computeSize(size, _maxSortedUsers);
    if (size != newSize) _heap.size = newSize;

    if (_formerValue != _newValue) {
        if (_newValue == 0) remove(_heap, _id, _formerValue);
        else if (_formerValue == 0) insert(_heap, _id, _newValue, _maxSortedUsers);
        else if (_formerValue < _newValue) increase(_heap, _id, _newValue, _maxSortedUsers);
        else decrease(_heap, _id, _newValue);
    }
}
```

### Recommendation:

While the contract is a `library` and thus input sanitization can be performed at the application level, from a security standpoint this reliance is ill-advised. Instead, we advise the `_formerValue` to be queried directly on the heap structure to guarantee all operations are performed safely regardless of end-user usage. If gas optimization is a concern here (as the data entry is already read by the caller), we advise a delta system to be used instead (i.e. specifying a decrease as `-5`) which would optimize the `if` conditionals and allow the entry to be read once at the heap level.

QGarchery:
The delta solution, just like loading the `_formerValue`, leads to one more (warm) load operation. Should we implement it to make sure the library is used as intended ?

MathisGD:
[...]

PR #49, opened 2022-07-06
Potential significant optimization (packing)

QGarchery:
Fixes 
- #35 

Discussion: can we really cast our values into uint96 without a risk of overflow ?

before

After changes:
after-tight-pack

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,6 @@
     "typescript": "4.5.2"
   },
   "dependencies": {
-    "@openzeppelin/contracts": "^4.7.0"
+    "@openzeppelin/contracts": "4.7.0"
   }
 }


MathisGD (approved):
2^96 does not seem unbreakable (~10 Bn $ in DAI). But, if it is, the issue is for this user, because he will not be at the right place in the heap. That is all, because we do the accounting of balances in the mappings in uint256.

test-foundry/TestStessHeapOrdering.t.sol
```diff
@@ -14,30 +14,30 @@ contract HeapStorage {
     function setUp() public {
         for (uint256 i = 0; i < TESTED_SIZE; i++) {
```
MerlinEgalite, line 15: perhaps just change here instead of casting below

MerlinEgalite (changes requested):
Not approving for now, I think we need to discuss it further 🙂

pakim249CAL:
This is not safe in the long term IMO, for reasons already stated previously. There is no guarantee that there won't be more assets added to underlying liquidity pools that would break things in a uint96 paradigm.

Rubilmax:
If I'm not mistaken, the maximum corresponds to 8e10 pool units / P2P units, which we store inside the data structure. So it does not correspond to 8e10 ERC20 18-decimals tokens.

With indexes ever growing, the amount of underlying ERC20 18-decimals tokens to reach such an amount would grow linearly with the index (and would currently be around 1e11 on small markets).

This change *may* be ok if the DAO never adds an ERC20 with more than 18 decimals or whose liquidity is too concentrated (or partial pause the market if it becomes so).

Morpho's operations include at least 2 DS operations (the position of the user). 800 gas on each DS operation would decrease Morpho's gas cost from at least 1600, maybe 4800.

Im not sure the risk is not worth the reward..

MathisGD:
No, you need a "safecast" to force it to revert.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/SafeCast.sol

MathisGD:
If we do that, it should be in the library directly, because really easy to forget.

MerlinEgalite:
> If we do that, it should be in the library directly, because really easy to forget.

Not sure we should do that in the lib. To me it's the responsibility of the contract integrating the lib to cast to the right type. Else you'll do an unnecessary casting if you're already using uint96

MathisGD:
> Not sure we should do that in the lib. To me it's the responsibility of the contract integrating the lib to cast to the right type. Else you'll do an unnecessary casting if you're already using uint96.

In our case, balances are uint256. So yes it is our responsibility, but it is also our lib, and we can do this in it IMO

QGarchery:
I agree with Mathis, and also `uint96` is rare so most of the time the value would be cast by the caller anyway.

Also I like the changes if it reverts when the value given is too big

MathisGD:
And yes I am also for the revert

contracts/HeapOrdering.sol
```diff
@@ -30,8 +30,8 @@ library HeapOrdering {
     function update(
         HeapArray storage _heap,
         address _id,
-        uint256 _formerValue,
-        uint256 _newValue,
+        uint96 _formerValue,
```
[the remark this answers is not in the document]
QGarchery: I think it is fine actually: morpho-contracts reference one particular commit. When we will decide to use the latest commit (that includes this change), we would also have to change how the `update` function is called

test-foundry/TestHeapOrdering.t.sol
```diff
@@ -401,16 +401,16 @@ contract TestHeapOrdering is DSTest {
     }
 
     function testInsertLast() public {
-        for (uint256 i; i < 10; i++) update(accounts[i], 0, NB_ACCOUNTS - i);
+        for (uint256 i; i < 10; i++) update(accounts[i], 0, uint96(NB_ACCOUNTS - i));
```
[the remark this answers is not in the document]
QGarchery: done

test-foundry/TestStessHeapOrdering.t.sol
QGarchery: done

QGarchery:
Safe casting added. Note that it would be cleaner to also do #43, because we would have only one casting to do.

QGarchery:
After safe casting 
still-the-one

package.json
```diff
@@ -42,5 +42,8 @@
     "solidity-coverage": "^0.7.16",
     "ts-node": "10.4.0",
     "typescript": "4.5.2"
+  },
+  "dependencies": {
+    "@openzeppelin/contracts": "^4.7.0"
```
MerlinEgalite, line 47: pin the dependency please

test-foundry/TestHeapOrdering.t.sol
```diff
@@ -503,4 +502,14 @@ contract TestHeapOrdering is DSTest {
 
         assertEq(heap.accounts[1].value, 25);
     }
+
+    function testOverflowNewValue() public {
+        hevm.expectRevert("SafeCast: value doesn't fit in 96 bits");
+        update(accounts[0], 0, uint256(type(uint128).max));
```
MerlinEgalite, line 508: why not `type(uint128).max + 1`?

test-foundry/TestStessHeapOrdering.t.sol
```diff
@@ -14,9 +14,9 @@ contract HeapStorage {
...
```
MerlinEgalite, line 15:
```suggestion
        for (uint96 i; i < TESTED_SIZE; i++) {
```

MerlinEgalite: as well for the `TESTED_SIZE`

MerlinEgalite (changes requested):
Please pin the dependency 🙂

test-foundry/TestStessHeapOrdering.t.sol
QGarchery: `uint96` is (mostly) not needed outside of the library now, because the casting is done inside

test-foundry/TestHeapOrdering.t.sol
QGarchery: It would revert with the change you propose. I can change it to `uint256(type(uint96).max)+1` but it is longer

package.json
QGarchery: Just added a commit. Is it fine now ?

test-foundry/TestHeapOrdering.t.sol
MathisGD: It is ok like this IMO

MathisGD (approved):

test-foundry/TestHeapOrdering.t.sol
MerlinEgalite: ok np

PR #51, opened 2022-07-07
Small improvements

QGarchery:


contracts/HeapOrdering.sol
```diff
@@ -123,13 +123,13 @@ library HeapOrdering {
     /// @param _heap The heap to modify.
     /// @param _rank The rank of the account to move.
     function shiftUp(HeapArray storage _heap, uint256 _rank) private {
-        Account memory initialAccount = getAccount(_heap, _rank);
-        uint256 initialValue = initialAccount.value;
-        while (_rank > 1 && initialValue > getAccount(_heap, _rank >> 1).value) {
+        Account memory accountToShift = getAccount(_heap, _rank);
+        uint256 valueToShift = accountToShift.value;
```
MathisGD, line 127: This variable saves 5 gas on average for `updateHeap` (just tried). We don't have to keep it (but we can)

Makefile
```diff
@@ -1,15 +1,19 @@
...
+export FOUNDRY_TEST=test-foundry
+
+install:
+	@git submodule update --init --recursive
+	@yarn
+
+test:
 	@echo Run all tests
-	@forge test -vvv -c test-foundry
+	@forge test -vvv
```
MathisGD, line 11: I think that for these simple repos, we can do without makefile (only foundry.toml). People that are used to foundry will prefer

QGarchery: Sure, but it does not cost much to also have the makefile. Plus it is really standard to have a makefile for different projects (not necessarily even related to blockchain)

MathisGD: Ok! Let's keep it

contracts/HeapOrdering.sol
MerlinEgalite: let's keep I'de say

MerlinEgalite (approved):
LGTM

PR #56, opened 2022-07-15
3-heap

QGarchery:
**Strongly** inspired by the MEP work.

This PR introduces a new kind of heap: the 3-heap, a heap with 3 child branches. Given an **index** `i`, the computations are as follows:
- the index of the left child is `3 * i + 1`
- the index of the middle child is `3 * i + 2`
- the index of the right child is `3 * i + 3`
- the index of the parent is `(i - 1) / 3` (given that `i` is not 0, the index of the root node)

We factorize the tests as much as possible, and add a pseudo-random test common to both the 2-heap and the 3-heap. This allows us to effectively compare their respective gas consumption. In the following, only the `update` function is significant, and the result are shown after running:
```
forge test -vvv -c test-foundry --match-test testHeapStructure --gas-report
```

Gas consumption of the 2-heap:
before2heap

Gas consumption of the 3-heap before:
before

Gas consumption of the 3-heap after simplifying the `shiftDown` function:
afterSimplifyShiftDown
Note that this commit is NOT meant to be an opti, but rather it is meant to simplify the code base to be able to maintain it more easily.

Gas consumption of the 3-heap after casting down to `uint96` (a packing opti):
afterCast


makcandrov (approved):

pakim249CAL:
Any reason you aren't using:
`using SafeCast for uint256`?

QGarchery:
> Any reason you aren't using: `using SafeCast for uint256`?

At first I thought we would have exactly one safecast so I did not bother. Now we have 2 (because we kept the `_formerValue` argument in `update`), but I'm still not sure if it is worth it. Do you feel like it would improve readability here ?

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,306 @@
...
+        uint256 newSize = computeSize(size, _maxSortedUsers);
+        if (size != newSize) _heap.size = newSize;
+
+        if (formerValue != newValue) {
+            if (newValue == 0) remove(_heap, _id, formerValue);
+            else if (formerValue == 0) insert(_heap, _id, newValue, _maxSortedUsers);
+            else if (formerValue < newValue) increase(_heap, _id, newValue, _maxSortedUsers);
+            else decrease(_heap, _id, newValue);
+        }
+    }
```
Rubilmax, line 56: These could be extracted to a base abstract contract (e.g. BaseHeap) and other methods could be defined virtual inside it and overridden in this implementation contract

Rubilmax: It's perhaps a matter of another PR though (TDD ; code, test, refactor)

Rubilmax (approved):

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,306 @@
...
+library ThreeHeapOrdering {
+    struct Account {
+        address id; // The address of the account.
+        uint96 value; // The value of the account.
+    }
+
+    struct HeapArray {
+        Account[] accounts; // All the accounts.
+        uint256 size; // The size of the heap portion of the structure, should be less than accounts length, the rest is an unordered array.
+        mapping(address => uint256) indexes; // A mapping from an address to a index in accounts.
```
MathisGD, line 15: In `HeapOrdering`, we use `indexOf`, which is a little bit more natural to me. What is the reason why you changed it here ?

QGarchery: Actually I changed it recently in `HeapOrdering` from `indexes` to `indexOf` (after @MerlinEgalite's review), and I forgot to also change it in this PR. I like `indexOf` better too, I'll change it

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,306 @@
...
+        uint256 _index
+    ) private {
+        uint256 size = _heap.size;
+        Account memory targetAccount = _accountToShift;
+        uint256 targetIndex = _index;
+        Account memory nextAccount;
+        uint256 nextIndex = _index * 3;
+        uint256 rightChildIndex;
+
+        for (;;) {
```
MerlinEgalite, line 125: not sure why not putting everything into a while

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,306 @@
...
+            rightChildIndex = nextIndex + 3;
```
MerlinEgalite, line 126: can be in an unchecked

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,306 @@
...
+    /// @param _index The index of the account to move.
+    function shiftDown(
+        HeapArray storage _heap,
+        Account memory _accountToShift,
+        uint256 _index
+    ) private {
+        uint256 size = _heap.size;
+        Account memory targetAccount = _accountToShift;
+        uint256 targetIndex = _index;
+        Account memory nextAccount;
```
MerlinEgalite, line 121: Does Initiating it here save gas instead of in the loop?

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,306 @@
...
+        uint256 nextIndex = _index * 3;
+        uint256 rightChildIndex;
+
+        for (;;) {
+            rightChildIndex = nextIndex + 3;
+            while (++nextIndex <= rightChildIndex && nextIndex < size) {
```
MerlinEgalite, line 127: so why not check set the `nextIndex` at `_index * 3 + 1` so you can write: 
`while (nextIndex < rightChildIndex && nextIndex < size) {`

I think it could save some gas. Especially if you uncheck `nextIndex = _index * 3 + 1`

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,306 @@
...
+        Account memory account = _heap.accounts[index];
+        if (account.id != _id) return 0;
+        else return account.value;
+    }
+
+    /// @notice Returns the address at the head of the `_heap`.
+    /// @param _heap The heap to get the head.
+    /// @return The address of the head.
+    function getHead(HeapArray storage _heap) internal view returns (address) {
+        if (_heap.accounts.length > 0) return _heap.accounts[0].id;
```
MerlinEgalite, line 271:
```suggestion
        if (_heap.accounts.length > 0) return _heap.accounts[ROOT].id;
```

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,306 @@
...
+    ) private {
+        uint256 index = _heap.indexes[_id];
+        delete _heap.indexes[_id];
+        uint256 accountsLength = _heap.accounts.length;
+
+        if (_heap.size == accountsLength) _heap.size--;
+        if (index == accountsLength - 1) {
+            _heap.accounts.pop();
+            return;
+        }
```
MerlinEgalite, line 235: I think this is less likely to happen, perhaps removing it save gas in general.

MerlinEgalite (changes requested):
Mainly open questions

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,306 @@
...
+    /// @param _heap The heap to modify.
+    /// @param _accountToShift The account to move.
+    /// @param _index The index of the account to move.
+    function shiftDown(
+        HeapArray storage _heap,
+        Account memory _accountToShift,
+        uint256 _index
+    ) private {
+        uint256 size = _heap.size;
+        Account memory targetAccount = _accountToShift;
```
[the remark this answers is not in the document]
QGarchery: There are 2 main changes that require changing the naming:
- we use a while loop to go through the three child. This was changed from the MEP solution because it greatly reduces the code size
- we pass the account to shift as an argument. It is a small optimization not in the HeapOrdering.

contracts/ThreeHeapOrdering.sol
QGarchery: I think it is the same thing

contracts/ThreeHeapOrdering.sol
QGarchery: It does not work: `nextIndex` is incremented in the while loop, whereas with your solution it is not

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,306 @@
...
+        HeapArray storage _heap,
+        Account memory _accountToShift,
+        uint256 _index
+    ) private {
+        Account memory parentAccount;
+        uint256 parentIndex;
+        while (
+            _index > ROOT &&
+            _accountToShift.value >
+            (parentAccount = _heap.accounts[parentIndex = (_index - 1) / 3]).value
```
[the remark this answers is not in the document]
QGarchery: Yes, it was a previous opti to not do the computation `(_index - 1) / 3` twice and we tested it

contracts/ThreeHeapOrdering.sol
QGarchery: No, it is slightly better in the loop. I'll change that

contracts/ThreeHeapOrdering.sol
QGarchery: We can't remove it easily because we pop `accounts` after that and assume that `index` is still in the range of `acounts`

contracts/ThreeHeapOrdering.sol
QGarchery: I won't do it in this PR, because there are many things that could be put into an unchecked

contracts/ThreeHeapOrdering.sol
MerlinEgalite: Ok updating my example so that `nextIndex` is updated but the former value is used: `while (nextIndex++ < rightChildIndex && nextIndex < size)`

contracts/ThreeHeapOrdering.sol
MerlinEgalite: But perhaps less clear idk, you could do:
```solidity
while (targetIndex != _index) {

    uint256 rightChildIndex = nextIndex + 3;
    while (++nextIndex <= rightChildIndex && nextIndex < size) {
            
        // ...
        
    }
    
    setAccount(_heap, _index, targetAccount);
    
    // ...
}
```

So that the condition is in the `while` statement. But not a blocker to me!

MerlinEgalite (approved):
Just made some suggestions but not a blocker to me though, changes look good!

contracts/ThreeHeapOrdering.sol
MerlinEgalite: Ok forget about it!

contracts/ThreeHeapOrdering.sol
MerlinEgalite: Ok not possible forget about it 😆

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,307 @@
...
+    /// @param _id The address of the account to decrease the amount.
+    /// @param _newValue The new value of the account.
+    function decrease(
+        HeapArray storage _heap,
+        address _id,
+        uint96 _newValue
+    ) private {
+        uint256 index = _heap.indexOf[_id];
+
+        if (index < (_heap.size - 1) / 3) shiftDown(_heap, Account(_id, _newValue), index);
```
MathisGD, line 189: I think that it is:
```suggestion
        if (index < _heap.size / 3) shiftDown(_heap, Account(_id, _newValue), index);
```

MathisGD:
Huge work, a pleasure to review ! Congrats @QGarchery 👌

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    /// @notice Moves an account down the heap until its value is greater than the ones of its children.
+    /// @dev This functions restores the invariant about the order of the values stored when the account to move is the only one with value smaller than what it should be.
+    /// @param _heap The heap to modify.
+    /// @param _accountToShift The account to move.
+    /// @param _index The index of the account to move.
+    function shiftDown(
+        HeapArray storage _heap,
+        Account memory _accountToShift,
```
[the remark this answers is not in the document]
QGarchery: I'm putting this in a new issue

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,307 @@
...
+
+        setAccount(_heap, _index, _accountToShift);
+    }
+
+    /// @notice Moves an account down the heap until its value is greater than the ones of its children.
+    /// @dev This functions restores the invariant about the order of the values stored when the account to move is the only one with value smaller than what it should be.
+    /// @param _heap The heap to modify.
+    /// @param _accountToShift The account to move.
+    /// @param _index The index of the account to move.
+    function shiftDown(
```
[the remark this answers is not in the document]
QGarchery: Yes ! It is slightly more costly but I find it significantly cleaner that way.

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,307 @@
...
+    /// @dev This functions restores the invariant about the order of the values stored when the account to move is the only one with value greater than what it should be.
+    /// @param _heap The heap to modify.
+    /// @param _accountToShift The account to move.
+    /// @param _index The index of the account to move.
+    function shiftUp(
+        HeapArray storage _heap,
+        Account memory _accountToShift,
+        uint256 _index
+    ) private {
+        Account memory parentAccount;
```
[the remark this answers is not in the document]
QGarchery: I would trust `testFullHeapSort` the most, but I'm not sure either. Since this is cleaner that way we will keep it like that I think

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,307 @@
...
+        HeapArray storage _heap,
+        Account memory _accountToShift,
+        uint256 _index
+    ) private {
+        uint256 size = _heap.size;
+        Account memory targetAccount = _accountToShift;
+        uint256 targetIndex = _index;
+        uint256 nextIndex = _index * 3;
+
+        for (;;) {
```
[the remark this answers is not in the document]
QGarchery: This is exactly the same compiled code. Let's replace the `for` loop with a `while` loop to harmonize with `shiftUp`. The second one I find it confusing to replace because it has some side effect

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,307 @@
...
+            uint256 rightChildIndex = nextIndex + 3;
```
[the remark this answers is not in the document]
QGarchery: Yes but not much actually. Since we don't really understand why this is happening (and I don't think it is a good use of our time to put variable declaration in and out of the loops), I won't change it. In the end I harmonized everything to put variable declarations inside of the loop when possible

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,307 @@
...
+        uint96 _value,
+        uint256 _maxSortedUsers
+    ) private {
+        // `_heap` cannot contain the 0 address.
+        if (_id == address(0)) revert AddressIsZero();
+
+        uint256 size = _heap.size;
+        uint256 accountsLength = _heap.accounts.length;
+
+        if (size < accountsLength) {
```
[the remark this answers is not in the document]
QGarchery: I like it ! This will be known as the Mathis optimization :)

MerlinEgalite (dismissed):
LGTM

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,307 @@
...
+    /// @param _value The value of the account to insert.
+    /// @param _maxSortedUsers The maximum size of the heap.
+    function insert(
+        HeapArray storage _heap,
+        address _id,
+        uint96 _value,
+        uint256 _maxSortedUsers
+    ) private {
+        // `_heap` cannot contain the 0 address.
+        if (_id == address(0)) revert AddressIsZero();
```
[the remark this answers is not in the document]
QGarchery: Agreed, although this feels more like an internal error (because we happen to use the address 0 to signify the absence of node). I clarified what are the sane input parameters in `update`

contracts/ThreeHeapOrdering.sol
MathisGD: You fixed this in #83

contracts/ThreeHeapOrdering.sol
```diff
@@ -163,12 +163,9 @@ library ThreeHeapOrdering {
         uint256 size = _heap.size;
         uint256 accountsLength = _heap.accounts.length;
```
MerlinEgalite, line 164: Does it cost more to remove this var?

QGarchery: Yes because we push just after and so we would replace it with `_heap.accounts.length -1`

contracts/ThreeHeapOrdering.sol
```diff
@@ -0,0 +1,307 @@
...
+        uint256 size = _heap.size;
+        uint256 newSize = computeSize(size, _maxSortedUsers);
+        if (size != newSize) _heap.size = newSize;
+
+        if (formerValue != newValue) {
+            if (newValue == 0) remove(_heap, _id, formerValue);
+            else if (formerValue == 0) insert(_heap, _id, newValue, _maxSortedUsers);
+            else if (formerValue < newValue) increase(_heap, _id, newValue, _maxSortedUsers);
+            else decrease(_heap, _id, newValue);
+        }
```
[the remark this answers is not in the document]
QGarchery: Done in #84

contracts/ThreeHeapOrdering.sol
MerlinEgalite: Hmm I'm not sure about the link you posted

QGarchery: > Hmm I'm not sure about the link you posted

Yes sorry, upload did not go through somehow

MerlinEgalite: Ok wanted to be sure thx

pakim249CAL (approved):

MathisGD (approved):

MerlinEgalite (approved):
LGTM

PR #68, opened 2022-09-22
Heap invariant testing

MathisGD:


MathisGD:
Question: should I remove the old `TestRandomHeap` ?

pakim249CAL:
The test run seems quite long? We should reduce the number of runs somehow, or not include this in our general testing commands/CI.

MathisGD:
> The test run seems quite long? We should reduce the number of runs somehow, or not include this in our general testing commands/CI.

Locally it was ok (~10min), but it seems that it is too heavy for the CI 😬

pakim249CAL (approved):

test-foundry/helpers/ConcreteThreeHeapOrdering.sol
```diff
@@ -13,6 +13,10 @@ contract ConcreteThreeHeapOrdering is ICommonHeapOrdering {
         return heap.accounts[_index].value;
     }
 
+    function accountsId(uint256 _index) external view returns (address) {
```
MerlinEgalite, line 16: `accountsValue` and `accountsId` are quite misleading since the returned value is linked to one single account, but clearly not a problem

test-foundry/TestInvariantHeapOrdering.t.sol
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: GNU AGPLv3
+pragma solidity ^0.8.0;
+
+import "ds-test/test.sol";
```
MerlinEgalite, line 4: Why this import? I think foundry allows you to directly import it now

MerlinEgalite (approved):
LGTM

test-foundry/helpers/ConcreteThreeHeapOrdering.sol
MathisGD: It has been done like this because of the name of the mapping. You would prefer `accountValue` and `accountId` right ?

test-foundry/TestInvariantHeapOrdering.t.sol
MathisGD: Wdym I don't understand. I need it for `DSTest`.

MerlinEgalite: Yes but just import import `"@forge-std/Test.sol";` then because `DSTest` is imported by it

MathisGD: done

test-foundry/helpers/ConcreteThreeHeapOrdering.sol
MerlinEgalite: As `indexOf`, we also can use `idAtIndex` and `valueAtIndex`

MathisGD: Idk `accountsId` and `accountsValue` were great too

MerlinEgalite (approved):
LGTM

MerlinEgalite (approved):
LGTM

pakim249CAL (approved):

test-foundry/helpers/TestRandomHeap.sol
```diff
@@ -1,87 +0,0 @@
...
-                else decrease();
-            }
-        }
-
-        uint256 lastValue = type(uint256).max;
-        uint256 newValue;
-        while ((newValue = removeHead()) != 0) {
-            require(newValue <= lastValue, "Elements are not given back in a decreasing order.");
-            lastValue = newValue;
-        }
```
QGarchery, line 85: I'm not sure we should remove this test, it is different from what we test in the invariant testing. I believe we could do something similar though

test-foundry/TestHeapOrderingInvariant.t.sol
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier: GNU AGPLv3
```
QGarchery, line 1: Maybe name it `InvariantHeapOrdering.t.sol` ? That way it is easier to match only invariant contracts (and it is cleaner IMO)

test-foundry/TestHeapOrderingInvariant.t.sol
```diff
@@ -0,0 +1,71 @@
...
+    // For all i in [[0, size]],
+    // value[i] >= value[2i + 1] and value[i] >= value[2i + 2]
+    function invariantHeap() public {
+        uint256 length = heap.length();
+
+        for (uint256 i; i < length; ++i) {
+            assertTrue((i * 2 + 1 >= length || i * 2 + 1 >= heap.size() || heap.accountsValue(i) >= heap.accountsValue(i * 2 + 1))); // prettier-ignore
+            assertTrue((i * 2 + 2 >= length || i * 2 + 2 >= heap.size() || heap.accountsValue(i) >= heap.accountsValue(i * 2 + 2))); // prettier-ignore
+        }
+    }
```
QGarchery, line 54: I think this is better than what we add before, so I would remove the function `verifyStructure` (which is not used right now)

QGarchery (changes requested):
Awesome work ! A few things to change I think

test-foundry/TestHeapOrderingInvariant.t.sol
MathisGD: I changed from this, because I found that it was a mess in my files explorer. But you raise a good point.

test-foundry/TestHeapOrderingInvariant.t.sol
MathisGD: Yes nice catch

test-foundry/helpers/TestRandomHeap.sol
MathisGD: Oh yes, this was not intended. I'm adding it back.

MathisGD: I will do, when the Heap would have been modernized a little bit (using indexes instead of ranks, etc)

test-foundry/TestHeapOrderingInvariant.t.sol
MathisGD: Btw, I think that the `Test` prefix everywhere is useless

QGarchery: > because I found that it was a mess in my files explorer.

Why do you think that ?

test-foundry/helpers/TestRandomHeap.sol
MathisGD: Yes I did it
https://github.com/morpho-dao/morpho-data-structures/issues/77

test-foundry/TestHeapOrderingInvariant.t.sol
MathisGD: now I have 

```
test
  ∟ TestHeapOrdering.t.sol
  ∟ TestHeapOrderingInvariant.t.sol
  ∟ TestThreeHeapOrdering.t.sol
  ∟ TestThreeHeapOrderingInvariant.t.sol
```

But that's a detail, I can change back if you want

QGarchery: Ok, let's keep it like that. Maybe one day we will remove the prefix `Test` and all will be good

QGarchery (approved):

Issue #63, opened 2022-08-08
Possible optimization in heap size


QGarchery:
Found by @MathisGD.

Currently the heap size is changing at each `insert`, `remove`, and `decrease` operation. When it reaches `_maxSortedUsers`, it is divided by 2. The informal goal is to distribute high liquidity evenly on the branches of the heap, in order to maximize the total liquidity that is in the heap.

One other idea would be distribute the incoming liquidity "randomly" on the last nodes before `_maxSortedUsers`. This means that the size of the heap would only change up to `_maxSortedUsers` and then stay the same, potentially saving one `SSTORE` for each operation. Another potential benefit would be that the introduced randomness would mostly prevent worst case scenarios for the incoming liquidity.
Note that there is no real need for the randomness here to be truly random, and we could use a pseudo-random function instead.

MerlinEgalite:
What's the state of this issue?

MathisGD:
It has to be implemented ! I started something in #76, but it was far from being ready. I might go back to it when I will have a little bit more bandwidth.

PR #73, opened 2022-10-10
perf: Pack heap size + array length

moodlezoup:
Replaces the `HeapArray.accounts` array with a "virtual array" consisting of a mapping and a `uint128` length. This lets us pack the array length and heap size into the same storage slot. Note that this trick also works for 3-heap

## Gas Before 
Screen Shot 2022-10-10 at 2 28 18 PM

## Gas After
Screen Shot 2022-10-10 at 2 44 14 PM

### DIFF
--- a/contracts/HeapOrdering.sol
+++ b/contracts/HeapOrdering.sol
@@ -10,8 +10,9 @@ library HeapOrdering {
     }
 
     struct HeapArray {
-        Account[] accounts; // All the accounts.
-        uint256 size; // The size of the heap portion of the structure, should be less than accounts length, the rest is an unordered array.
+        mapping(uint256 => Account) accounts; // All the accounts.
+        uint128 arrayLength; // The length of the array represented by the `accounts` mapping.
+        uint128 size; // The size of the heap portion of the structure, should be less than `arrayLength`, the rest is an unordered array.
         mapping(address => uint256) indexOf; // A mapping from an address to an index in accounts. From index i, the parent index is (i-1)/2, the left child index is 2*i+1 and the right child index is 2*i+2.
     }
 
@@ -38,13 +39,13 @@ library HeapOrdering {
         address _id,
         uint256 _formerValue,
         uint256 _newValue,
-        uint256 _maxSortedUsers
+        uint128 _maxSortedUsers
     ) internal {
         uint96 formerValue = SafeCast.toUint96(_formerValue);
         uint96 newValue = SafeCast.toUint96(_newValue);
 
-        uint256 size = _heap.size;
-        uint256 newSize = computeSize(size, _maxSortedUsers);
+        uint128 size = _heap.size;
+        uint128 newSize = computeSize(size, _maxSortedUsers);
         if (size != newSize) _heap.size = newSize;
 
         if (formerValue != newValue) {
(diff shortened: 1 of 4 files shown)


QGarchery (commented):
That is an awesome PR, and a really cool trick to know about !

There are 2 things to consider that maybe do not make it worth it to merge. The first one is that arrays in solidity may get optimized later, and the current solution would cost less gas. Right now the generated code does not make use of the fact that array elements are contiguous, and the hash to the first array position gets computed everytime the array is accessed. This means that a lot of gas could be saved when we iterate over the array, like we do in `shiftDown` and `shiftUp`.

The other point is that there is a further optimization, see #63, that does not require updating the size as often (and in fact never at all after the heap is initialized). The reason why we introduced the `size` parameter is to distribute incoming liquidity evenly accross the leaves. Indeed, the size is divided by 2 to add liquidity at the left leaf, and the size increment shifts new liquidity to the right.

contracts/HeapOrdering.sol
```diff
@@ -237,12 +239,16 @@ library HeapOrdering {
...
         // Swap the last account and the account to remove, then pop it.
         swap(_heap, index, accountsLength - 1);
-        if (_heap.size == accountsLength) _heap.size--;
-        _heap.accounts.pop();
+        unchecked {
+            // accountsLength - 1 is performed in the line above, so these cannot underflow.
+            delete _heap.accounts[accountsLength - 1];
+            _heap.arrayLength = accountsLength - 1;
+            if (_heap.size == accountsLength) _heap.size = accountsLength - 1;
+        }
```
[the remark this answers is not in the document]
MerlinEgalite: I agree that we could replace L242 by something like that:
```solidity
uint128 newArrayLength = _heap.arrayLength - 1;
```
and use this variable instead.

contracts/HeapOrdering.sol
```diff
@@ -168,18 +169,19 @@ library HeapOrdering {
...
     ) private {
         // `_heap` cannot contain the 0 address.
         if (_id == address(0)) revert AddressIsZero();
 
         // Put the account at the end of accounts.
-        uint256 accountsLength = _heap.accounts.length;
-        _heap.accounts.push(Account(_id, _value));
+        uint128 accountsLength = _heap.arrayLength;
+        _heap.accounts[accountsLength] = Account(_id, _value);
+        _heap.arrayLength = accountsLength + 1;
```
MerlinEgalite, line 180:
```suggestion
        uint128 accountsLength = _heap.arrayLength++; // Increase _heap.arrayLength while returning the former value.
        _heap.accounts[accountsLength] = Account(_id, _value);
```

MerlinEgalite (changes requested):
Some gas improvement suggestions but overall LGTM, nice work!

contracts/HeapOrdering.sol
moodlezoup: this actually increases the gas a little bit 😞

MerlinEgalite: woww really? Note sure to get why....

Btw gas is tested at 200 optimizer runs, we might need to test at an higher number of runs.

contracts/HeapOrdering.sol
```diff
@@ -126,7 +127,7 @@ library HeapOrdering {
     /// @param _heap The heap to modify.
     /// @param _index The index of the account to move.
     function shiftDown(HeapArray storage _heap, uint256 _index) private {
-        uint256 size = _heap.size;
+        uint128 size = _heap.size;
```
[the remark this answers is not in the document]
MathisGD: I did it in #75, which I wanted to open on top of this one but unfortunately I'm not able to

QGarchery:
Closing this PR as changes are also in #75

PR #74, opened 2022-10-12
Opti: avoid `_heap.size` reads

MathisGD:
Small optimization of `HeapOrdering`. See each commit as a separated optimization, we can keep each of them separately. My preferred one is the first one, the second and the first one adding a little bit of code.

See the results for `updateHeap` in `TestRandomHeapOrdering`:
Gas before: 21922
Gas after: 21811

### DIFF
--- a/contracts/HeapOrdering.sol
+++ b/contracts/HeapOrdering.sol
@@ -48,10 +48,11 @@ library HeapOrdering {
         if (size != newSize) _heap.size = newSize;
 
         if (formerValue != newValue) {
-            if (newValue == 0) remove(_heap, _id, formerValue);
-            else if (formerValue == 0) insert(_heap, _id, newValue, _maxSortedUsers);
-            else if (formerValue < newValue) increase(_heap, _id, newValue, _maxSortedUsers);
-            else decrease(_heap, _id, newValue);
+            if (newValue == 0) remove(_heap, _id, newSize, formerValue);
+            else if (formerValue == 0) insert(_heap, _id, newSize, newValue, _maxSortedUsers);
+            else if (formerValue < newValue)
+                increase(_heap, _id, newSize, newValue, _maxSortedUsers);
+            else decrease(_heap, _id, newSize, newValue);
         }
     }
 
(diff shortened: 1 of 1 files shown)


contracts/HeapOrdering.sol
```diff
@@ -234,20 +239,24 @@ library HeapOrdering {
...
         uint256 index = _heap.indexOf[_id];
         uint256 accountsLength = _heap.accounts.length;
 
         // Swap the last account and the account to remove, then pop it.
         swap(_heap, index, accountsLength - 1);
-        if (_heap.size == accountsLength) _heap.size--;
+        if (_size == accountsLength) {
+            _size--;
+            _heap.size = _size;
+        }
```
QGarchery, line 253:
```suggestion
        if (_size == accountsLength) _heap.size = --_size;
```

QGarchery (approved):
Let's go !

contracts/HeapOrdering.sol
```diff
@@ -244,12 +244,15 @@ library HeapOrdering {
 
         // Swap the last account and the account to remove, then pop it.
         swap(_heap, index, accountsLength - 1);
-        if (_size == accountsLength) _heap.size = _size - 1;
+        if (_size == accountsLength) {
+            _size--;
+            _heap.size = _size;
```
MerlinEgalite: and if you try just: `_heap.size = _size--;`?

And in an unchecked?

MerlinEgalite (changes requested):
Can you try it please? Else LGTM!

contracts/HeapOrdering.sol
MathisGD: Yes it has already been changed ;-)

MerlinEgalite: wut? I'm not sure what happened 😅

MerlinEgalite (approved):
LGTM

contracts/HeapOrdering.sol
MathisGD: I think that you started a review then I changed it and then you submitted your review :-)

PR #75, opened 2022-10-12
Pack heap size + array length

MathisGD:
Extension of #73 by @moodlezoup.

Replaces the HeapArray.accounts array with a "virtual array" consisting of a mapping and a uint128 length. This lets us pack the array length and heap size into the same storage slot.

I just added the use of `uint256` as much as possible, preventing Solidity to go back and forth between `uint128` and `uint256`. I also removed useless and costly `else`.

image


contracts/HeapOrdering.sol
```diff
@@ -237,12 +239,17 @@ library HeapOrdering {
...
+        uint256 accountsLength = _heap.arrayLength;
 
         // Swap the last account and the account to remove, then pop it.
-        swap(_heap, index, accountsLength - 1);
-        if (_heap.size == accountsLength) _heap.size--;
-        _heap.accounts.pop();
+        uint256 newArrayLength = accountsLength - 1;
+        swap(_heap, index, newArrayLength);
+        unchecked {
+            // accountsLength - 1 is performed in the line above, so these cannot underflow.
```
moodlezoup, line 248: I forgot to delete this comment

contracts/HeapOrdering.sol
```diff
@@ -174,15 +175,16 @@ library HeapOrdering {
         if (_id == address(0)) revert AddressIsZero();
 
         // Put the account at the end of accounts.
-        uint256 accountsLength = _heap.accounts.length;
-        _heap.accounts.push(Account(_id, _value));
+        uint256 accountsLength = _heap.arrayLength;
+        _heap.accounts[accountsLength] = Account(_id, _value);
+        _heap.arrayLength = uint128(accountsLength + 1);
```
moodlezoup, line 180: I guess it's fine to not use SafeCast because these cannot realistically overflow?

moodlezoup (approved):
thanks! feel free to close my PR in favor of this one

MerlinEgalite:
Can I review after the fix of the conflicts?

MathisGD:
Rebased and updated @MerlinEgalite

src/HeapOrdering.sol
```diff
@@ -248,13 +250,15 @@ library HeapOrdering {
...
-        uint256 accountsLength = _heap.accounts.length;
+        uint256 accountsLength = _heap.arrayLength;
 
         // Swap the last account and the account to remove, then pop it.
-        swap(_heap, index, accountsLength - 1);
-        if (_size == accountsLength) _heap.size = --_size;
-        _heap.accounts.pop();
+        uint256 newArrayLength = accountsLength - 1;
+        swap(_heap, index, newArrayLength);
+        delete _heap.accounts[newArrayLength];
```
MerlinEgalite, line 258: perhaps add a comment that it is possible because we are taking the length in sync at any time. Because `delete` does not update `accounts`'s length

QGarchery (commented):
I don't think we should merge this PR, for the reasons explained in [this comment](https://github.com/morpho-dao/morpho-data-structures/pull/73#pullrequestreview-1136941488). It would change the data-structure to improve gas consumption, when doing #63 would be better in that regard, and also better because it would spread liquidity evenly. So if we want to change the heap ordering I think we should go all the way and implement #63

src/HeapOrdering.sol
MerlinEgalite: To me this is not clear enough. It's not obvious that `delete` is not updating the array's length. It could for other reasons that `arrayLength` is updated eg the access is cheaper than `array.length`

src/HeapOrdering.sol
```diff
@@ -248,13 +250,15 @@ library HeapOrdering {
...
         delete _heap.indexOf[_id];
+        _heap.arrayLength = uint128(newArrayLength);
+        if (_heap.size == accountsLength) _heap.size = uint128(--_size);
```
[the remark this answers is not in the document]
MathisGD: aie rebasing issue

src/HeapOrdering.sol
MathisGD: > It's not obvious that delete is not updating the array's length

There is no longer any array

MathisGD: is it clearer like this ?

MerlinEgalite:
> I don't think we should merge this PR, for the reasons explained in [this comment](https://github.com/morpho-dao/morpho-data-structures/pull/73#pullrequestreview-1136941488). It would change the data-structure to improve gas consumption, when doing #63 would be better in that regard, and also better because it would spread liquidity evenly. So if we want to change the heap ordering I think we should go all the way and implement #63

given this comment, should we review? @MathisGD @QGarchery

QGarchery (changes requested):
> given this comment, should we review?

Probably not, unless there are other arguments in favor of merging it I think we should close this PR for now

PR #78, opened 2022-10-29
Perf three heap (uncheckatoooooor)

MerlinEgalite:
This is just a suggestion. It might be possible to do the same on the "normal" heap.

### Reference
Screenshot 2022-10-29 at 10 31 26

### first commit
Screenshot 2022-10-29 at 14 44 43

### Second commit
Screenshot 2022-10-29 at 14 48 58

### Third commit
Screenshot 2022-10-29 at 14 50 10

### DIFF
--- a/contracts/ThreeHeapOrdering.sol
+++ b/contracts/ThreeHeapOrdering.sol
@@ -189,8 +189,11 @@ library ThreeHeapOrdering {
     ) private {
         uint256 index = _heap.indexOf[_id];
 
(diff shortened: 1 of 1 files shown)


MathisGD:
nooo 😢

MathisGD:
It is not my PR so I will let Quentin choose :-)

QGarchery:
I think that if we want to optimize a lot the next step would be to write parts of the whole thing in Yul, so I'm closing this PR for now

MerlinEgalite:
As you want

PR #82, opened 2022-11-11
chore: use foundry.toml instead of Makefile

MathisGD:
Use `foundry.toml` instead of Makefile


foundry.toml
```diff
@@ -0,0 +1,3 @@
+[profile.default]
```
MerlinEgalite, line 1: I think we should set more parameters like solc version

MerlinEgalite: Yes np, but I think it should be especially if we're doing gas snapshot to compare

MathisGD: We are not doing this here. So should I do it ? If yes, with which version ?

foundry.toml
```diff
@@ -0,0 +1,3 @@
...
+src = "contracts"
+test = "test-foundry"
```
QGarchery, line 3: Please insert a newline at the end of the file. There is a setting in VSCode to not have to worry about this: it is called "Files: Insert Final Newline"

QGarchery (approved):
Yes please no more makefile :)

Makefile
```diff
@@ -1,19 +0,0 @@
...
-
-export FOUNDRY_TEST=test-foundry
-
-install:
-	@git submodule update --init --recursive
-	@yarn
-
-test:
-	@echo Run all tests
-	@forge test -vvv
```
QGarchery, line 11: Both of those are advertised in the README, we can either keep them or change the doc

foundry.toml
MathisGD: Not sure to understand why, but ok why not.

Makefile
MerlinEgalite: It would be nice to also add in the doc how to assess gas consumption. There's nothing written written on this for now

MathisGD: Do you feel that it is useful to tell people how to run tests when the setup is the one everyone uses ?

MathisGD: Same question for you @MerlinEgalite. Everything is written in the foundry book already !

foundry.toml
MerlinEgalite: I thought you wanted to add gas snapshot on this repo. If not then you can forget about it

Makefile
QGarchery: We can the test section from the README, or keep part of the makefile. Either is ok for me

QGarchery: (there is also the install section)

MerlinEgalite: The REAMDE works for me with a mention to the test to check to compare gas is enough

QGarchery (approved):

Makefile
MathisGD: >The REAMDE works for me with a mention to the test to check to compare gas is enough

What do you mean ? There are test to compare the gas cost ? with what ?

MathisGD: I updated the README

MerlinEgalite: I think @QGarchery you were using a randomized test for gas comparison no?

QGarchery: Yes `testFullHeapSort` is the best I feel. There is also `testHeapStructure` but a large part of the gas consumption is due to the checking so it's less reliable

MathisGD: Ok I see. I dont think that it is worth having this in the README.

MerlinEgalite (approved):
LGTM

Makefile
MerlinEgalite: Ok as you want, it's not clear to me how an outside collaborator will know how to compare gas though

PR #83, opened 2022-11-11
Fix shift guard in decrease

QGarchery:
### Problem
We want to skip `shiftDown` when there are no elements below it in the heap part of the data-structure. 

### Solution

Just check if the first son is in the heap part of the data-structure. This corresponds to `leftMostChildIndex(index) < _heap.size`.

For the 3-heap, this is simply `3 * index + 1 < _heap.size`.

For the 2-heap, this can be checked more efficiently with `index < _heap.size / 2`. This comes from the following general theorem:
$\forall m ~n \in \mathbb{N}. m < \frac{n}{2} \Leftrightarrow 2m + 1 < n$

### Other checks
The following ways to check (for the 3-heap) don't work. 
1. `index < (_heap.size - 1) / 3`
2. `index < _heap.size / 3`

See the added test `testDecreaseGuardLimit` for a counter-example.

### DIFF
--- a/contracts/HeapOrdering.sol
+++ b/contracts/HeapOrdering.sol
@@ -198,7 +198,7 @@ library HeapOrdering {
         uint256 index = _heap.indexOf[_id];
         _heap.accounts[index].value = _newValue;
 
-        // We only need to restore the invariant if the account is a node in the heap
+        // We only need to take care of sorting if there are nodes below in the heap.
         if (index < _heap.size >> 1) shiftDown(_heap, index);
     }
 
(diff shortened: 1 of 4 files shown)


pakim249CAL (approved):

MathisGD (approved):
LGTM 

Just one thing: tests pass when I replace the < by <= (so when the left child is the first account of the non-heap). Maybe it is worth adding a test for this ?

QGarchery:
> LGTM
> 
> Just one thing: tests pass when I replace the < by <= (so when the left child is the first account of the non-heap). Maybe it is worth adding a test for this ?

Hmm not sure how to test that since it has the correct behavior, it is just less efficient

MathisGD:
With <=, isn't there a case we are not shifting down but we don't need to ?

MerlinEgalite (approved):
LGTM

QGarchery:
> With <=, isn't there a case we are not shifting down but we don't need to ?

We shift down more with the `<=` but that's not an issue because shifting would stop immediatly

PR #84, opened 2022-11-11
Memoization of size in the 3-heap

QGarchery:
Also reorders arguments to have heap then size (the initial state), and the the specific arguments of the functions like `_id` or `_value`.

### DIFF
--- a/contracts/ThreeHeapOrdering.sol
+++ b/contracts/ThreeHeapOrdering.sol
@@ -220,7 +220,7 @@ library ThreeHeapOrdering {
     /// @notice Removes an account in the `_heap`.
     /// @dev Only call when this function `_id` is in the `_heap` with value `_removedValue`.
     /// @param _heap The computed heap to modify.
-    /// @param _size The size of the heap.
+    /// @param _size The computed size of the heap.
     /// @param _id The address of the account to remove.
     /// @param _removedValue The value of the account to remove.
     function remove(
--- a/contracts/ThreeHeapOrdering.sol
+++ b/contracts/ThreeHeapOrdering.sol
@@ -48,10 +48,11 @@ library ThreeHeapOrdering {
         if (size != newSize) _heap.size = newSize;
 
         if (formerValue != newValue) {
-            if (newValue == 0) remove(_heap, _id, formerValue);
-            else if (formerValue == 0) insert(_heap, _id, newValue, _maxSortedUsers);
-            else if (formerValue < newValue) increase(_heap, _id, newValue, _maxSortedUsers);
-            else decrease(_heap, _id, newValue);
+            if (newValue == 0) remove(_heap, newSize, _id, formerValue);
+            else if (formerValue == 0) insert(_heap, newSize, _id, newValue, _maxSortedUsers);
+            else if (formerValue < newValue)
+                increase(_heap, newSize, _id, newValue, _maxSortedUsers);
+            else decrease(_heap, newSize, _id, newValue);
         }
     }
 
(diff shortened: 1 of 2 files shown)


contracts/ThreeHeapOrdering.sol
```diff
@@ -148,83 +150,90 @@ library ThreeHeapOrdering {
...
-        if (index < (_heap.size - 1) / 3) shiftDown(_heap, Account(_id, _newValue), index);
+        // We only need to take care of sorting if there are nodes below in the heap.
+        if (3 * index + 1 < _size) shiftDown(_heap, _size, Account(_id, _newValue), index);
         else _heap.accounts[index].value = _newValue;
     }
 
     /// @notice Increases the amount of an account in the `_heap`.
     /// @dev Only call this function when `_id` is in the `_heap` with a smaller value than `_newValue`.
     /// @param _heap The heap to modify.
+    /// @param _size The computed size of the heap.
```
MathisGD, line 199: I would put "the size of the heap" everywhere

MerlinEgalite (approved):
It would be nice to have a gas comparison since it's difficult to know if it saves gas or not. (I think so but not sure)

MathisGD (approved):
LGTM. This would be a great use of [Transient storage](https://eips.ethereum.org/EIPS/eip-1153), in the future.

contracts/ThreeHeapOrdering.sol
QGarchery: Computed is nice because it reminds the reader of the `computeSize` function. This is not the size that is stored in the DLL so I think it is worth adding this to differentiate

MathisGD:
Capture d’écran 2022-11-11 à 18 30 43

contracts/ThreeHeapOrdering.sol
MathisGD: I think that for `increase` at least, the size is the one stored in the heap

MathisGD: And I think that it is always the case

MerlinEgalite:
> LGTM. This would be a great use of [Transient storage](https://eips.ethereum.org/EIPS/eip-1153), in the future.

Yep, however, I'm not sure you can do it with a struct and a lib...

contracts/ThreeHeapOrdering.sol
QGarchery: > I think that for increase at least, the size is the one stored in the heap

No it's not the one stored in the heap because `_maxSortedUser` can change from transaction to transaction

MathisGD: But when we call update we store the new size right ?

pakim249CAL (approved):

contracts/ThreeHeapOrdering.sol
QGarchery: Oh ok I see what you mean. I was thinking about the size before the transaction. Adding computed insists on the fact that it is not the size stored in the heap before the transaction, so I feel like it's worth it to add it. If you are not convinced we can go back, it's also ok with me

MathisGD: Ok I see. Then can you add it everywhere please ?

MathisGD: Otherwise replace by « new »

QGarchery: Ok done. I kept computed because new can be misleading in the case of insert

MathisGD: Ok perfect

MathisGD (approved):

PR #91, opened 2022-12-10
Correct testing behavior for safe cast

MerlinEgalite:
Forme tests where testing for the wrong value (`uint128` instead of `uint96`)


MathisGD (approved):

Rubilmax (approved):
This seems like a great test case for fuzzing

MerlinEgalite:
> This seems like a great test case for fuzzing

yep! will push that

MathisGD (approved):
Note that you could also have fuzzed the address

Rubilmax (approved):

PR #94, opened 2022-12-14
Propagate morpho-v1 setup

Rubilmax:
- The goal is to have contracts under `src/` and forge tests under `test/`, so I propagated `morpho-v1`'s setup
- The `Makefile` seems to have been removed in favor of a pure `forge` setup. However, it still enables some handy shortcuts to the dev, increasing the devexp (at least to me). Can we have it back?
- I tried enabling viaIR, but it takes too much time when compiling test contracts... It'd be better to compare gas costs with IR compilation though


pakim249CAL (approved):

MathisGD (approved):
Not a fan of Makefiles for solidity but not a blocker

MerlinEgalite (approved):

Makefile
```diff
@@ -0,0 +1,31 @@
+-include .env.local
+.EXPORT_ALL_VARIABLES:
+MAKEFLAGS += --no-print-directory
+
```
QGarchery, line 4:
```suggestion
-include .env.local
.EXPORT_ALL_VARIABLES:
MAKEFLAGS += --no-print-directory

install:
        yarn
	foundryup
	git submodule update --init --recursive
```
So that calling `make` from a fresh repository allows you to call other forge commands. 

Actually this is the only target that I would be adding to the makefile. As @MathisGD, this is not a blocker for me if we keep the other targets

QGarchery (changes requested):
Looks good, one change requested

PR #95, opened 2022-12-15
Logarithmic buckets

Morphojo:
poc

### DIFF
--- a/contracts/LogarithmicBuckets.sol
+++ b/contracts/LogarithmicBuckets.sol
@@ -15,7 +15,7 @@ library LogarithmicBuckets {
 
     /// CONSTANTS ///
 
-    uint256 private constant LOGBASE = 4;
+    uint256 private constant LOGBASE = 2;
 
     /// ERRORS ///
 
--- a/contracts/LogarithmicBuckets.sol
+++ b/contracts/LogarithmicBuckets.sol
@@ -1,0 +1,124 @@
+// SPDX-License-Identifier: GNU AGPLv3
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/utils/math/SafeCast.sol";
+import "./DoubleLinkedList.sol";
+
+library LogarithmicBuckets {
+    using DoubleLinkedList for DoubleLinkedList.List;
+
+    struct BucketList {
+        mapping(uint256 => DoubleLinkedList.List) lists; // All the accounts.
+        mapping(address => uint256) indexOf;
+        uint256 maxIndex;
+    }
+
+    /// CONSTANTS ///
+
+    uint256 private constant LOGBASE = 4;
+
+    /// ERRORS ///
+
+    /// @notice Thrown when the address is zero at insertion.
+    error AddressIsZero();
+
+    /// INTERNAL ///
+
+    /// @notice Updates an account in the `_heap`.
+    /// @dev Only call this function when `_id` is in the `_heap` with value `_formerValue` or when `_id` is not in the `_heap` with `_formerValue` equal to 0.
+    function update(
+        BucketList storage _buckets,
+        address _id,
+        uint256 _formerValue,
+        uint256 _newValue
+    ) internal {
+        uint96 formerValue = SafeCast.toUint96(_formerValue);
+        uint96 newValue = SafeCast.toUint96(_newValue);
+
+        if (formerValue != newValue) {
+            if (formerValue != 0) remove(_buckets, _id);
+            if (newValue != 0) insert(_buckets, _id, newValue);
+        }
+    }
+
+    /// PRIVATE ///
+
+    function insert(
+        BucketList storage _buckets,
+        address _id,
+        uint96 _value
+    ) internal {
+        // `_buckets` cannot contain the 0 address.
+        if (_id == address(0)) revert AddressIsZero();
+        uint256 bucketIndex = computeBucketIndex(_value);
+        _buckets.lists[bucketIndex].insertTail(_id, _value);
+        _buckets.indexOf[_id] = bucketIndex;
+        if (bucketIndex > _buckets.maxIndex) _buckets.maxIndex = bucketIndex;
+    }
+
+    /// @notice Removes an account in the `_buckets`.
+    /// @param _buckets The buckets to modify.
+    /// @param _id The address of the account to remove.
+    function remove(BucketList storage _buckets, address _id) internal {
+        uint256 index = _buckets.indexOf[_id];
+        // Swap the last account and the account to remove, then pop it.
+        _buckets.lists[index].remove(_id);
+        delete _buckets.indexOf[_id];
+
+        while (_buckets.lists[index].head == address(0) && index > 0) {
+            index -= 1;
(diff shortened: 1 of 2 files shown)


contracts/LogarithmicBuckets.sol
```diff
@@ -0,0 +1,124 @@
...
+
+    /// @notice Compute the bucket index.
+    /// @param _value The value of the index to compute.
+    function computeBucketIndex(uint96 _value) private pure returns (uint256) {
+        for (uint256 i = 0; i < 128; i++) {
+            if (_value < LOGBASE**i) {
+                return i;
+            }
+        }
+        return 128;
```
MerlinEgalite, line 82: We might consider more efficient ways to do it later on, such as [this](https://graphics.stanford.edu/~seander/bithacks.html#IntegerLogDeBruijn) presented [here](https://ethereum.stackexchange.com/questions/8086/logarithm-math-operation-in-solidity).

contracts/LogarithmicBuckets.sol
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /// GETTERS ///
```
MerlinEgalite, line 85:
```suggestion
    /// GETTERS ///
    
```

contracts/LogarithmicBuckets.sol
```diff
@@ -0,0 +1,124 @@
...
+    /// @notice Returns the value of the account linked to `_id`.
+    /// @param _buckets The buckets to search in.
+    /// @param _id The address of the account.
+    /// @return value The value of the account.
+    function getValueOf(BucketList storage _buckets, address _id) internal view returns (uint256) {
+        uint256 index = _buckets.indexOf[_id];
+        DoubleLinkedList.Account memory account = _buckets.lists[index].accounts[_id];
+        return account.value;
```
MerlinEgalite, line 93: Might save gas

```suggestion
        return _buckets.lists[_buckets.indexOf[_id]].accounts[_id].value;
```

contracts/LogarithmicBuckets.sol
```diff
@@ -0,0 +1,124 @@
...
+        uint96 formerValue = SafeCast.toUint96(_formerValue);
+        uint96 newValue = SafeCast.toUint96(_newValue);
+
+        if (formerValue != newValue) {
+            if (formerValue != 0) remove(_buckets, _id);
+            if (newValue != 0) insert(_buckets, _id, newValue);
+        }
+    }
+
+    /// PRIVATE ///
```
MerlinEgalite, line 44: private and internal functions are mixed below

contracts/LogarithmicBuckets.sol
```diff
@@ -0,0 +1,124 @@
...
+        if (bucketIndex > _buckets.maxIndex) _buckets.maxIndex = bucketIndex;
+    }
+
+    /// @notice Removes an account in the `_buckets`.
+    /// @param _buckets The buckets to modify.
+    /// @param _id The address of the account to remove.
+    function remove(BucketList storage _buckets, address _id) internal {
+        uint256 index = _buckets.indexOf[_id];
+        // Swap the last account and the account to remove, then pop it.
+        _buckets.lists[index].remove(_id);
```
MerlinEgalite, line 65: Perhaps add a comment here saying that if `_id` does not exist it reverts

contracts/LogarithmicBuckets.sol
```diff
@@ -0,0 +1,124 @@
...
+        delete _buckets.indexOf[_id];
+
+        while (_buckets.lists[index].head == address(0) && index > 0) {
+            index -= 1;
+            _buckets.maxIndex = index;
```
MerlinEgalite, line 70: Should be stored in a temporary variable to save gas and update the storage variable only at the end.

contracts/LogarithmicBuckets.sol
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /// @notice Returns the value of the account linked to `_id`.
+    /// @param _buckets The buckets to search in.
+    /// @return value The value of the account.
+    function getMaxIndex(BucketList storage _buckets) internal view returns (uint256) {
+        return _buckets.maxIndex;
+    }
+
+    /// @notice Returns the address at the head of the `_heap`.
```
MerlinEgalite, line 103: Comment must be updated

contracts/LogarithmicBuckets.sol
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    /// @notice Compute the bucket index.
+    /// @param _value The value of the index to compute.
+    function computeBucketIndex(uint96 _value) private pure returns (uint256) {
+        for (uint256 i = 0; i < 128; i++) {
+            if (_value < LOGBASE**i) {
+                return i;
+            }
```
MerlinEgalite, line 80: I think it's more this to check no?
 
```suggestion
            if (LOGBASE**i <=_value  && _value < LOGBASE**i) {
                return i;
            }
```

MerlinEgalite (changes requested):
Nice work!

contracts/LogarithmicBuckets.sol
Morphojo: you mean i-1 no? but since the i's are increasing only 1 check is necessary.

contracts/LogarithmicBuckets.sol
```diff
@@ -0,0 +1,124 @@
...
+        }
+        return 128;
+    }
```
QGarchery, line 83: There are versions of log2 that are extremely gas efficient. A C implementation can be found here: https://graphics.stanford.edu/~seander/bithacks.html#IntegerLogDeBruijn
The explanation is that you can find De Bruijn sequences which are sequences that have no repeating substring of a given length. This means that multiplying this sequence by a power of 2 is an injective function, so you can precompute the results (off-chain) and map them back to a power of 2. You can get log2 of a number this way by first finding the nearest power of 2 of your number
An implementation in solidity can be found here (not sure if it is correct):
https://ethereum.stackexchange.com/a/30168/51843

QGarchery (commented):
Awesome work ! :tada: 

I think we need to optimize this structure a lot before it can be used (I added comments for that). Another idea is to link buckets between them, so we don't have to increase one by one like it is done in the `getHead` function right now

contracts/LogarithmicBuckets.sol
MerlinEgalite: No `i` because you're starting from 0, no?

QGarchery: You can't be both greater than and strictly lower than `LOGBASE**i`, so I don't think the code you suggest works @MerlinEgalite. Also, since we already checked that the value is not strictly smaller that a certain threshold in the previous iterations, I think it would be redundant to check it. In the end the code @Morphojo wrote seems to be working

MerlinEgalite: ah yes forgot the +1

MerlinEgalite: Yes it works my bad

Issue #54, opened 2022-07-12
Migrate last tests to `forge`


MathisGD:
And remove hardhat from the repo

MathisGD:
Don't forget to remove useless contracts

PR #110, opened 2023-01-02
Make imports relative

Rubilmax:
This PR implements solution 2) described in https://github.com/morpho-dao/morpho-utils/issues/79

Please read the issue & the discussion first!!

Sometimes, you have to make ugly choices to keep backward compatibility...

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,12 +8,12 @@
   "scripts": {
     "test": "npx hardhat test test-ts/*.ts",
     "coverage": "npx hardhat coverage",
-    "lint": "yarn run lint:sol && yarn run lint:js",
+    "lint": "yarn run lint:sol && yarn run lint:ts",
     "lint:ts": "eslint . --ext .ts",
     "lint:sol": "solhint 'contracts/**/*.sol'",
-    "fix": "yarn fix:sol && yarn fix:js",
+    "fix": "yarn fix:sol && yarn fix:ts",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "fix": "yarn fix:sol && yarn fix:js",
     "fix:ts": "prettier --config .prettierrc.json './**/*.ts' --write",
     "fix:sol": "prettier --write 'contracts/**/*.sol'",
-    "size": "yarn run hardhat size-contracts",
+    "size": "npx hardhat size-contracts",
```
MathisGD, line 17: why not using forge ?
```suggestion
    "size": "forge build --sizes",
```

MathisGD (approved):

QGarchery (approved):

MerlinEgalite (approved):
Note that OZ lib version is updated to the `4.8.0` at the same time

pakim249CAL (approved):

package.json
Rubilmax: I shouldn't have changed this since it's out of the scope of this PR
I think one could argue that we shouldn't mix `hardhat` running on `nodejs` with `forge` which requires a separate installation, but I also think that `hardhat` should disappear progressively from this repository

MathisGD: Yes, poke #54

PR #117, opened 2023-04-25
Refresh `BasicHeap` library and testing suite

makcandrov:
This PR is an attempt to refresh the `BasicHeap` library [`Heap.sol`](https://github.com/morpho-dao/morpho-data-structures/blob/main/src/Heap.sol) which previously used the old design of heaps with ranks instead of indexes, and the testing suite was quite old as well.

# Changes in the library

The change mainly involves shifting all the indexes to the left (`index = rank - 1`). The consequences are as follows:
- Indexes now start at 0 (previously, ranks started at 1).
- The children of the account at `index` are `(2 * index + 1, 2 * index + 2)` (previously, the children of the account at `rank` werre `(2 * rank, 2 * rank + 1)`).
- We can no longer check if an account is in the heap simply by comparing index to zero. Previously, we had `rank != 0` <=> the account is in the heap. Now we also need to check if `_heap.accounts[0].id == _id`.

I took the liberty of allowing the insertion of an account with a null value, which was previously forbidden (it threw the error `WrongValue()`). It makes sense in [`HeapOrdering.sol`](https://github.com/morpho-dao/morpho-data-structures/blob/main/src/HeapOrdering.sol) to do so because we insert or remove by discriminating on whether this value is zero or not. In [`Heap.sol`](https://github.com/morpho-dao/morpho-data-structures/blob/main/src/Heap.sol), the user specifies whether they want to insert or remove from the heap, so disallowing a zero value doesn't make sense.

# Changes in the testing suite

The testing suite used pseudo-randomness to manually implement fuzzing. This PR changes it to use the invariants tests proposed by Foundry. All the contracts related to pseudo-randomness are deleted.

# Left to do

While refreshing this library, I came across many incoherent things and remnants of old designs and testing tricks. In order to refresh and harmonize all the files from this repository, I think that some other changes are needed, but I believe that they deserve separate PRs. They are:

- [ ] All the testing files use the extension `.t.sol`, which we don't use anymore in recent repositories.  
- [ ] All th files use the old syntax of imports (`import "./TestCommonHeapOrdering.t.sol";` instead or `import {TestCommonHeapOrdering} from "./TestCommonHeapOrdering.t.sol";`).
- [ ] Maybe it would be better to gather all the invariant tests files in a separate folder under test.
- [ ] Because we are testing libraries, we cannot make a relevant gas report if we directly call the internal functions on the heap. That's why some tests use a "concrete heap" contract (see [`ConcreteHeapOrdering.sol`](https://github.com/morpho-dao/morpho-data-structures/blob/main/test/helpers/ConcreteHeapOrdering.sol)) that is essentially the "contract version" of the library. However, not all tests use this (see [`TestHeap.t.sol`](https://github.com/morpho-dao/morpho-data-structures/blob/main/test/TestHeap.t.sol)), and some of them add functions that are not in the library in the first place (see `accountsValue`, `accountsId` and `accountsId` from [`ConcreteHeapOrdering.sol`](https://github.com/morpho-dao/morpho-data-structures/blob/main/test/helpers/ConcreteHeapOrdering.sol) and [`ConcreteThreeHeapOrdering.sol`](https://github.com/morpho-dao/morpho-data-structures/blob/main/test/helpers/ConcreteThreeHeapOrdering.sol)). In my opinion, all tests should use a "contract version" of the library but with strictly the same functions.
- [ ] The testing suite generally lacks fuzzing.
- [ ] Many tests check that the storage is exactly what we expect after a series of operations (see [`testRemoveShiftUp`](https://github.com/morpho-dao/morpho-data-structures/blob/ba30cb23052c14dfb399ce87109d06fd868022e1/test/TestHeap.t.sol#L294-L319)). I don't think these are good tests because they are highly dependent on the implementation of the heap. For the same series of operations, many different orders in the list could be correct. For example if we insert the values `0`, `1`, and `2` in the heap, the lists `[2, 0, 1]` and `[2, 1, 0]` would be correct. However, the testing suite would fail in one of these situations.

Please let me know what you think about the changes I made and the changes I recommend doing! :)


QGarchery (changes requested):
Thanks for this PR and the detailed explanation ! I did not have the time to verify everything yet, but it sounds good so far. I think there are conflicts with the sync PR, so we should rebase right now to resolve them

src/Heap.sol
```diff
@@ -9,9 +9,13 @@ library BasicHeap {
 
     struct Heap {
         Account[] accounts; // All the accounts.
-        mapping(address => uint256) ranks; // A mapping from an address to a rank in accounts. Beware: ranks are shifted by one compared to indexes, so the first rank is 1 and not 0.
+        mapping(address => uint256) indexOf; // A mapping from an address to a rank in accounts. Beware: ranks are shifted by one compared to indexes, so the first rank is 1 and not 0.
```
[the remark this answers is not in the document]
pakim249CAL: You  mean the comment, right?

makcandrov:
@MerlinEgalite  I just wanted to hear your thoughts on it before starting anything, but yes, I'll create the issues later

MerlinEgalite:
I've created this discussion: https://github.com/morpho-dao/morpho-data-structures/discussions/122 to answer and discuss you questions @makcandrov to avoid polluting this PR.

src/Heap.sol
```diff
@@ -32,209 +36,171 @@ library BasicHeap {
...
-        Heap storage _heap,
-        address _id,
-        uint256 _newValue
-    ) internal {
-        uint256 rank = _heap.ranks[_id];
-        if (rank == 0) revert AccountDoesNotExist();
-        uint256 oldValue = getAccount(_heap, rank).value;
+    function increase(Heap storage _heap, address _id, uint256 _newValue) internal {
+        uint256 index = _heap.indexOf[_id];
+        if (index == 0 && (_heap.accounts.length == 0 || _heap.accounts[0].id != _id)) revert AccountDoesNotExist();
```
MerlinEgalite, line 73: Maybe we should also store the length of accounts in this function?

src/Heap.sol
```diff
@@ -32,209 +36,171 @@ library BasicHeap {
...
     function remove(Heap storage _heap, address _id) internal {
-        uint256 rank = _heap.ranks[_id];
-        if (rank == 0) revert AccountDoesNotExist();
-        uint256 removedValue = getAccount(_heap, rank).value;
-        uint256 size = getSize(_heap);
+        uint256 index = _heap.indexOf[_id];
+        uint256 accountsLength = _heap.accounts.length;
 
-        if (rank == size) {
+        if (index == 0 && (accountsLength == 0 || _heap.accounts[0].id != _id)) revert AccountDoesNotExist();
```
MerlinEgalite, line 89: Perhaps it's worth internalizing the logic. It's used at least 3 times

src/Heap.sol
```diff
@@ -32,209 +36,171 @@ library BasicHeap {
...
+            while (
+                _index > ROOT
+                    && _accountToShift.value > (parentAccount = _heap.accounts[parentIndex = (_index - 1) >> 1]).value
+            ) {
+                setAccount(_heap, parentAccount, _index);
+                _index = parentIndex;
+            }
         }
-        setAccount(_heap, _rank, initialAccount);
+        setAccount(_heap, _accountToShift, _index);
```
MerlinEgalite, line 139:
```suggestion

        setAccount(_heap, _accountToShift, _index);
```

src/Heap.sol
```diff
@@ -32,209 +36,171 @@ library BasicHeap {
...
     /// VIEW ///
 
-    /// @notice Returns the size of the `_heap`.
+    /// @notice Returns the number of users in the `_heap`.
     /// @param _heap The heap parameter.
-    /// @return The size of the heap.
+    /// @return The length of the heap.
     function getSize(Heap storage _heap) internal view returns (uint256) {
         return _heap.accounts.length;
     }
```
MerlinEgalite, line 187: Should we use this function instead of `_heap.accounts.length` each time?

src/Heap.sol
```diff
@@ -32,209 +36,171 @@ library BasicHeap {
...
+        // At this point, childIndex (resp. childIndex+1) is the index of the left (resp. right) child.
+
+        while (childIndex < _size) {
+            Account memory childToSwap = _heap.accounts[childIndex];
+
+            // Find the child with largest value.
+            unchecked {
+                rightChildIndex = childIndex + 1; // This cannot overflow because childIndex < size.
+            }
+            if (rightChildIndex < _size) {
```
MerlinEgalite, line 160:
```suggestion
            }
            
            if (rightChildIndex < _size) {
```

src/Heap.sol
makcandrov: I'm not sure I understand
you're proposing to use `getSize` instead of `_heap.accounts.length` throughout the entire code?

makcandrov:
@MerlinEgalite I added the function `containsAccount` to internalize the logic. I took the opportunity to add an external getter.

src/Heap.sol
makcandrov: Since it's not used after, I don't think it's useful

src/Heap.sol
MerlinEgalite: Yes that's the idea. Just to make the code more concise, but not sure it's a good idea tbh

src/Heap.sol
```diff
@@ -32,209 +36,199 @@ library BasicHeap {
...
     /// @param _heap The heap to search in.
     /// @param _id The address of the account.
     /// @return The value of the account.
     function getValueOf(Heap storage _heap, address _id) internal view returns (uint256) {
-        uint256 rank = _heap.ranks[_id];
-        if (rank == 0) return 0;
-        else return getAccount(_heap, rank).value;
+        uint256 index;
+
+        if (!containsAccount(_heap, index = _heap.indexOf[_id], _heap.accounts.length, _id)) return 0;
```
pakim249CAL, line 223: Why does index need to be created and assigned here?

src/Heap.sol
```diff
@@ -243,28 +237,40 @@ library BasicHeap {
...
     /// @param _id The address to get the right child.
     /// @return The address of the right child.
     function getRightChild(Heap storage _heap, address _id) internal view returns (address) {
-        uint256 rank = _heap.ranks[_id] * 2 + 1;
-        if (rank != 1 && rank <= getSize(_heap)) return getAccount(_heap, rank).id;
-        else return address(0);
+        uint256 index;
+        uint256 accountsLength;
+
+        if (!containsAccount(_heap, index = _heap.indexOf[_id], accountsLength = _heap.accounts.length, _id)) {
```
pakim249CAL, line 269: I think these could be if, else if, else

src/Heap.sol
```diff
@@ -243,28 +237,40 @@ library BasicHeap {
     /// @param _id The address to get the parent.
     /// @return The address of the parent.
     function getParent(Heap storage _heap, address _id) internal view returns (address) {
-        uint256 rank = _heap.ranks[_id] / 2;
-        if (rank != 0) return getAccount(_heap, rank).id;
-        else return address(0);
+        uint256 index;
+        if ((index = _heap.indexOf[_id]) == 0) return address(0);
```
pakim249CAL, line 241: Can be if else?

src/Heap.sol
```diff
@@ -32,209 +36,199 @@ library BasicHeap {
...
+    /// @param _index The index to search for the account in.
+    /// @param _accountsLength The length of the `_heap` accounts array.
+    /// @param _id The address of the account to search for.
+    /// @return True if the account exists in the `_heap`, false otherwise.
+    function containsAccount(Heap storage _heap, uint256 _index, uint256 _accountsLength, address _id)
+        private
+        view
+        returns (bool)
+    {
+        if (_index != 0) return true;
```
pakim249CAL, line 192: This could be if, else if, else

src/Heap.sol
```diff
@@ -32,209 +36,199 @@ library BasicHeap {
...
+                if (rightChild.value > childToSwap.value) {
+                    childToSwap = rightChild;
+                    childIndex = rightChildIndex;
+                }
+            }
+
+            if (childToSwap.value > _accountToShift.value) {
+                setAccount(_heap, childToSwap, _index);
+                _index = childIndex;
+                childIndex = (childIndex << 1) + 1;
```
pakim249CAL, line 171: I believe this cannot overflow, which means that it would be safe to have this entire code block unchecked.

src/Heap.sol
```diff
@@ -32,209 +36,199 @@ library BasicHeap {
...
+
+        delete _heap.indexOf[_id];
 
-        if (rank == size) {
+        if (index == accountsLength - 1) {
             _heap.accounts.pop();
-            delete _heap.ranks[_id];
         } else {
-            swap(_heap, rank, size);
+            Account memory lastAccount = _heap.accounts[accountsLength - 1];
```
pakim249CAL, line 94: Can't we use index instead of accountsLength - 1?

src/Heap.sol
```diff
@@ -32,209 +36,199 @@ library BasicHeap {
...
     /// @param _id The address of the account to remove.
     function remove(Heap storage _heap, address _id) internal {
-        uint256 rank = _heap.ranks[_id];
-        if (rank == 0) revert AccountDoesNotExist();
-        uint256 removedValue = getAccount(_heap, rank).value;
-        uint256 size = getSize(_heap);
+        uint256 index = _heap.indexOf[_id];
+        uint256 accountsLength = _heap.accounts.length;
+
+        if ((!containsAccount(_heap, index, accountsLength, _id))) revert AccountDoesNotExist();
```
pakim249CAL, line 87: This line ensures that accountsLength > 0. Therefore this entire function can run unchecked.

src/Heap.sol
```diff
@@ -32,209 +36,199 @@ library BasicHeap {
...
+
+        delete _heap.indexOf[_id];
 
-        if (rank == size) {
+        if (index == accountsLength - 1) {
```
pakim249CAL, line 91: If we can make an assumption that accountsLength > 0, then this entire function can be unchecked.

makcandrov: Yes it's safe to assume that `accountsLength > 0` because at this point we know that the account is in the heap.
I made the suggested modification, the code is more optimized but less readable

Tristan22400 (dismissed):

src/Heap.sol
```diff
@@ -55,10 +53,9 @@ library BasicHeap {
     /// @param _newValue The new value of the account.
     function decrease(Heap storage _heap, address _id, uint256 _newValue) internal {
         uint256 index = _heap.indexOf[_id];
-        if (index == 0 && (_heap.accounts.length == 0 || _heap.accounts[0].id != _id)) revert AccountDoesNotExist();
 
-        uint256 oldValue = _heap.accounts[index].value;
-        if (_newValue >= oldValue) revert WrongValue();
+        if ((!containsAccount(_heap, index, _heap.accounts.length, _id))) revert AccountDoesNotExist();
```
MerlinEgalite, line 57: I think there is an unnecessary parenthesis, no?

MerlinEgalite (approved):

src/Heap.sol
```diff
@@ -36,209 +40,202 @@ library BasicHeap {
...
-        if (rank == 0) revert AccountDoesNotExist();
-        uint256 oldValue = getAccount(_heap, rank).value;
-        if (_newValue >= oldValue || _newValue == 0) revert WrongValue();
-
-        setAccountValue(_heap, rank, _newValue);
-        shiftDown(_heap, rank);
+    function decrease(Heap storage _heap, address _id, uint256 _newValue) internal {
+        uint256 index = _heap.indexOf[_id];
+
+        if ((!containsAccount(_heap, index, _heap.accounts.length, _id))) revert AccountDoesNotExist();
```
MerlinEgalite, line 61:
```suggestion
        if (!containsAccount(_heap, index, _heap.accounts.length, _id)) revert AccountDoesNotExist();
```

There's still the unnecessary parenthesis no?

MerlinEgalite (changes requested):

src/Heap.sol
makcandrov: you're right mb

Tristan22400 (approved):

src/Heap.sol
```diff
@@ -36,209 +40,202 @@ library BasicHeap {
...
+        private
+        view
+        returns (bool)
+    {
+        if (_index != 0) {
+            return true;
+        } else if (_accountsLength != 0) {
+            return _heap.accounts[0].id == _id;
+        } else {
+            return false;
```
QGarchery, line 202:
```suggestion
        return _index != 0 || _heap.accounts[0].id == _id;
```
This is because if the length is 0 then the first account should be empty. We could even have something like
```suggestion
        return _heap.accounts[_index].id == _id;
```
but I'm guessing that it's less efficient

src/Heap.sol
```diff
@@ -36,209 +40,202 @@ library BasicHeap {
...
-            setAccount(_heap, _rank, getAccount(_heap, _rank / 2));
-            _rank /= 2;
+    /// @param _size The size of the heap.
+    /// @param _accountToShift The account to move.
+    /// @param _index The index of the account to move.
+    function shiftDown(Heap storage _heap, uint256 _size, Account memory _accountToShift, uint256 _index) private {
+        uint256 valueToShift = _accountToShift.value;
+        uint256 childIndex = (_index << 1) + 1;
+        uint256 rightChildIndex;
+        // At this point, childIndex (resp. childIndex+1) is the index of the left (resp. right) child.
```
QGarchery, line 155: Actually if we put this comment at the very beginning of the loop it explains the loop invariant, and it makes it clearer why we set `childIndex` that way at the end of the loop

src/Heap.sol
```diff
@@ -36,209 +40,202 @@ library BasicHeap {
...
     /// @param _heap The heap to search in.
     /// @param _id The address of the account.
     /// @return The value of the account.
     function getValueOf(Heap storage _heap, address _id) internal view returns (uint256) {
-        uint256 rank = _heap.ranks[_id];
-        if (rank == 0) return 0;
-        else return getAccount(_heap, rank).value;
+        uint256 index;
+
+        if (!containsAccount(_heap, index = _heap.indexOf[_id], _heap.accounts.length, _id)) return 0;
```
QGarchery, line 230:
```suggestion
        uint256 index = _heap.indexOf[_id];

        if (!containsAccount(_heap, index, _heap.accounts.length, _id)) return 0;
```
clearer IMO

Actually can we just return `_heap.accounts[index].value` and not do the `if` ?

MerlinEgalite (approved):
Nice!

QGarchery (commented):
A few changes to make but it looks nice :)

src/Heap.sol
```diff
@@ -36,209 +40,202 @@ library BasicHeap {
...
     }
 
     /// VIEW ///
 
-    /// @notice Returns the size of the `_heap`.
+    /// @notice Returns the number of users in the `_heap`.
     /// @param _heap The heap parameter.
-    /// @return The size of the heap.
+    /// @return The length of the heap.
     function getSize(Heap storage _heap) internal view returns (uint256) {
```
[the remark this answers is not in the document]
makcandrov: perhaps just `length` ?

src/Heap.sol
makcandrov: We can't implement the last suggestion because it would revert if the length is zero.

Regarding the first suggestion, I initially tried that approach, but I was too afraid that the compiler might not always short-circuit boolean expressions (if the compiler evaluates both expressions in `_index != 0 || _heap.accounts[0].id == _id`, it can cause the function to revert). That's why I opted for the current implementation. I believe it makes the code more understandable and doesn't increase the gas cost.

src/Heap.sol
```diff
@@ -36,209 +40,202 @@ library BasicHeap {
...
-                childRank *= 2;
-            } else break;
+    /// @notice Checks if an account with the given address `_id` exists in the `_heap`.
+    /// @dev The parameters `_index`, `_accountsLength`, and `_id` must be coherent.
+    /// @param _heap The heap to search in.
+    /// @param _index The index to search for the account in.
+    /// @param _accountsLength The length of the `_heap` accounts array.
+    /// @param _id The address of the account to search for.
+    /// @return True if the account exists in the `_heap`, false otherwise.
+    function containsAccount(Heap storage _heap, uint256 _index, uint256 _accountsLength, address _id)
```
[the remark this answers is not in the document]
makcandrov: should I apply the same convention to `shiftDown`, `shiftUp` and `setAccount`?

test/TestHeapInvariant.t.sol
```diff
@@ -0,0 +1,136 @@
...
+        if (accountsUsed.length == 0) {
+            return;
+        }
+        index %= accountsUsed.length;
+        address account = accountsUsed[index];
+        uint256 accountValue = getValueOf(account);
+        if (accountValue == type(uint256).max) {
+            return;
+        }
+        increase(account, amount > accountValue ? amount : amount % (type(uint256).max - accountValue) + accountValue);
```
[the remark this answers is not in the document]
makcandrov: Yes, but overall, the entire testing suite should be refreshed (see #122). Perhaps we should reserve that for another upcoming PR.

src/Heap.sol
makcandrov: `accounts` is an array not a mapping, it would revert

src/Heap.sol
makcandrov: Since you're both in favor, I'm wrapping it inside an unchecked block.

MerlinEgalite (dismissed):

pakim249CAL (dismissed):

test/TestHeapInvariant.t.sol
```diff
@@ -0,0 +1,136 @@
...
+        return heap.accounts[_index].id;
+    }
+
+    function indexOf(address _id) external view returns (uint256) {
+        return heap.indexOf[_id];
+    }
+
+    /// Functions to fuzz ///
+
+    function insertCorrect(address account, uint256 amount) external {
```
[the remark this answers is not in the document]
makcandrov: The names `insertTrackUsed` and `insertTrackUsedAccounts` seem weird to me. Couldn't we think of anything else?

src/Heap.sol
makcandrov: @QGarchery

src/Heap.sol
makcandrov: Do we agree on not implementing this one?

test/TestHeapInvariant.t.sol
```diff
@@ -0,0 +1,136 @@
...
+        if (accountsUsed.length == 0) {
+            return;
+        }
+        index %= accountsUsed.length;
+        address account = accountsUsed[index];
+        uint256 accountValue = getValueOf(account);
+        if (accountValue == 0) {
+            return;
+        }
+        decrease(account, amount < accountValue ? amount : amount % accountValue);
```
Tristan22400, line 53: same remark as Quentin, you can use a bound here.

Tristan22400 (approved):

MerlinEgalite:
Let's merge IMO

makcandrov:
I have one last interrogation @MerlinEgalite , should we rename `getSize` for `getLength` or `length` (or not rename it at all)?

MerlinEgalite (approved):

pakim249CAL:
I think just length is appropriate. This would keep in line with pretty much any other data structure that is used out there.

Tristan22400 (approved):

src/Heap.sol
QGarchery: I'm just suggesting to move this comment 2 lines down, just below the line with the `while` keyword

src/Heap.sol
```diff
@@ -160,26 +160,26 @@ library BasicHeap {
             // Find the child with largest value.
             unchecked {
```
QGarchery, line 161: (optional) we could move this `unchecked` outside of the `while` loop. It should not change anything but maybe easier to read because it would not mix with the logic of the loop (and also we have done that for the other functions such as `_shiftUp`)

QGarchery (approved):

test/TestHeapInvariant.t.sol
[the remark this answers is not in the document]
QGarchery: I think it's fine to be very explicit in the tests as it's code that will not be reused. I would even be ok with `insertAndTrackUsedAccounts`

MerlinEgalite (dismissed):

MerlinEgalite (dismissed):

QGarchery (approved):

test/TestHeapInvariant.t.sol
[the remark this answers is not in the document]
makcandrov: I think I will rename these functions when refactoring the tests (#127) together with the functions from `MockHeapOrdering` & `ThreeMockHeapOrdering`. Let's not delay the merge any more

Tristan22400 (approved):

src/Heap.sol
```diff
@@ -36,209 +40,202 @@ library BasicHeap {
...
+    function insert(Heap storage _heap, address _id, uint256 _value) internal {
         if (_id == address(0)) revert AddressIsZero();
-        if (_heap.ranks[_id] != 0) revert AccountAlreadyInserted();
 
-        // Put the account at the end of the heap.
-        _heap.accounts.push(Account(_id, _value));
-        uint256 size = getSize(_heap);
-        _heap.ranks[_id] = size;
+        uint256 accountsLength = _heap.accounts.length;
+        if (_containsAccount(_heap, _heap.indexOf[_id], accountsLength, _id)) revert AccountAlreadyInserted();
```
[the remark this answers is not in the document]
makcandrov: In heap orderings, the check is implicitely done [here](https://github.com/morpho-org/morpho-data-structures/blob/47000974e8ef9cce097b86d2326de33f19320ad1/src/HeapOrdering.sol#L50-L56). Because the only entry point is the function `update`, nothing needs to be checked after.

PR #119, opened 2023-04-25
Sync repo

QGarchery:


MathisGD:
Sadly, we loose the reference to all the PRs that made all this work (merge commits point to PRs of this repo). Should we at least keep the private repo as a public archive for references ?

MathisGD:
As I understood it, the strategy for synchronizing the two repo requires rolling back main at some point. I'm wondering if it would not be better to merge/rebase the changes of the private repo on public/main instead.

QGarchery:
> As I understood it, the strategy for synchronizing the two repo requires rolling back main at some point. I'm wondering if it would not be better to merge/rebase the changes of the private repo on public/main instead.

I tried it and there is no conflict, but it does not mean that the added PRs are safe to rebase on. Doing so would prevent us to review the changes, because it would create one big PR with all the changes instead of 2 small PRs to review again

MathisGD:
If you "merge" we don't need to review everything right ?

MerlinEgalite (dismissed):

QGarchery:
> If you "merge" we don't need to review everything right ?

I think you do, merging is not safe. I can easily create two branches that are both fine by themselves, but merging introduces a bug, even if there is no conflict between the two

Tristan22400:
Is it okay to have unverified commit at the beginning of the PR ?

MathisGD:
Yes it is ok @Tristan22400

Tristan22400 (dismissed):

pakim249CAL (approved):

makcandrov (dismissed):

Issue #65, opened 2022-08-16
Abstract satellite data in the Heap


QGarchery:
We should also provide a way to abstract the satellite data in the structure, that way it can be reusable

_Originally posted by @QGarchery in https://github.com/morpho-dao/morpho-data-structures/issues/23#issuecomment-1186575374_

Tristan22400:
I understand that it adds complexity for a non-substancial benefit. I will close the issue and the PR if you agree.

PR #124, opened 2023-04-27
fix: abstract-satellite-data

Tristan22400:
Fix : https://github.com/morpho-dao/morpho-data-structures/issues/65
### Standardize Heap Code 
The goal of this PR is to generalize the Heap's code to enable everyone to use the data-structure by making it more general.

### DIFF
--- a/src/Heap.sol
+++ b/src/Heap.sol
@@ -2,9 +2,14 @@
 pragma solidity ^0.8.0;
 
 library BasicHeap {
+    struct RandomStruct {
+        uint256 value; // The value of the account.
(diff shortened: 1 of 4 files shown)


MathisGD:
Agree with Quentin that the fact that the dev would have to modify the library didn't really fixed the issue. I'm against merging this PR as is. 

And thinking about it a little bit, I don't think that there is a big need to generalize the data stored in the DS. What's interesting in the DS are the implem and gas tricks but not a potential ultra general aspect. Anyone can fork this repo and adapt the DSs to fit their needs if needed.

PR #129, opened 2023-05-03
Refactor to remove Head and Tail address

Tristan22400:
Fixes : https://github.com/morpho-org/morpho-data-structures/issues/85

### DIFF
--- a/src/DoubleLinkedList.sol
+++ b/src/DoubleLinkedList.sol
@@ -105,38 +105,11 @@ library DoubleLinkedList {
                 ++numberOfIterations;
             }
         }
-
-        // Account is not the new tail.
-        if (numberOfIterations < maxIterations && next != address(0)) {
-            // Account is the new head.
-            if (next == list.accounts[address(0)].next) {
-                list.accounts[id] = Account({prev: address(0), next: next, value: value});
-                list.accounts[address(0)].next = id;
-                list.accounts[next].prev = id;
-            }
-            // Account is not the new head.
-            else {
-                address prev = list.accounts[next].prev;
-                list.accounts[id] = Account({prev: prev, next: next, value: value});
-                list.accounts[prev].next = id;
-                list.accounts[next].prev = id;
-            }
+        if (numberOfIterations == maxIterations) {
+            next = address(0);
         }
-        // Account is the new tail.
-        else {
-            // Account is the new head.
-            if (list.accounts[address(0)].next == address(0)) {
-                list.accounts[id] = Account({prev: address(0), next: address(0), value: value});
-                list.accounts[address(0)].next = id;
-                list.accounts[address(0)].prev = id;
-            }
-            // Account is not the new head.
-            else {
-                address tail = list.accounts[address(0)].prev;
-                list.accounts[id] = Account({prev: tail, next: address(0), value: value});
-                list.accounts[tail].next = id;
-                list.accounts[address(0)].prev = id;
-            }
-        }
+        list.accounts[id] = Account(list.accounts[next].prev, next, value);
+        list.accounts[list.accounts[next].prev].next = id;
+        list.accounts[next].prev = id;
     }
 }
--- a/src/DoubleLinkedList.sol
+++ b/src/DoubleLinkedList.sol
@@ -77,6 +77,7 @@ library DoubleLinkedList {
     /// @param id The address of the account.
     function remove(List storage list, address id) internal {
         Account memory account = list.accounts[id];
+        if (id == address(0)) revert AddressIsZero();
         if (account.value == 0) revert AccountDoesNotExist();
 
         list.accounts[account.prev].next = account.next;
--- a/src/DoubleLinkedList.sol
+++ b/src/DoubleLinkedList.sol
@@ -16,8 +16,6 @@ library DoubleLinkedList {
 
     struct List {
         mapping(address => Account) accounts;
-        address head;
-        address tail;
     }
 
     /* ERRORS */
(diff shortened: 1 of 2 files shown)


src/DoubleLinkedList.sol
```diff
@@ -79,12 +77,11 @@ library DoubleLinkedList {
     /// @param id The address of the account.
     function remove(List storage list, address id) internal {
         Account memory account = list.accounts[id];
+        if (id == address(0)) revert AddressIsZero();
```
MathisGD, line 80: useless as the value of account zero cannot be something else than zero

MathisGD (changes requested):
I think that `insertSorted` can be completely refactored with this new design. See https://github.com/morpho-org/morpho-v1/pull/1255

src/DoubleLinkedList.sol
Tristan22400: I don't think it is useless as if you remove the account of address(0). You open the loop and you cannot find tail and head easily.

Tristan22400: I will do the changes as expected for the insertSorted function.

Tristan22400: yes you are right.

QGarchery:
While I love the idea, it would also require a big refactor of the verification done with Certora. Is it worth to invest some time into it ? @morpho-org/onchain-team

MerlinEgalite:
> While I love the idea, it would also require a big refactor of the verification done with Certora. Is it worth to invest some time into it ? @morpho-org/onchain-team

If it would take you 1 day or something I'm against and we can forget about it

QGarchery:
> > While I love the idea, it would also require a big refactor of the verification done with Certora. Is it worth to invest some time into it ? @morpho-org/onchain-team
> 
> If it would take you 1 day or something I'm against and we can forget about it

Yes probably one or two days to do it: I know well how it works now but it still requires to refactor almost all the invariants

src/DoubleLinkedList.sol
MathisGD: I think that it was an error

src/DoubleLinkedList.sol
```diff
@@ -100,46 +97,19 @@ library DoubleLinkedList {
...
         uint256 numberOfIterations;
-        address next = list.head; // If not added at the end of the list `id` will be inserted before `next`.
+        address next = list.accounts[address(0)].next; // If not added at the end of the list `id` will be inserted before `next`.
 
         while (numberOfIterations < maxIterations && next != address(0) && list.accounts[next].value >= value) {
             next = list.accounts[next].next;
             unchecked {
                 ++numberOfIterations;
             }
         }
```
MathisGD, line 107: this should work too, idk what's best 
```suggestion
        unchecked {
            while (numberOfIterations++ < maxIterations && next != address(0) && list.accounts[next].value >= value) {
                next = list.accounts[next].next;
            }
        }
```

src/DoubleLinkedList.sol
```diff
@@ -100,46 +97,19 @@ library DoubleLinkedList {
...
-            // Account is not the new head.
-            else {
-                address tail = list.tail;
-                list.accounts[id] = Account({prev: tail, next: address(0), value: value});
-                list.accounts[tail].next = id;
-                list.tail = id;
-            }
+        if (numberOfIterations == maxIterations) {
+            next = address(0);
         }
```
MathisGD, line 110: this works too, but maybe less readable
```suggestion
        assembly {
            next := mul(next, iszero(eq(numberOfIterations, maxIterations)))
        }
```

src/DoubleLinkedList.sol
makcandrov: To minimize storage reads (most heavy operation), I would have written it like that:

```solidity
uint256 numberOfIterations;
address prevId = address(0);
address nextId = list.accounts[address(0)].next; // If not added at the end of the list `id` will be inserted before `next`.
Account storage nextAccount = list.accounts[nextId];

unchecked {
    while (true) {
        if (++numberOfIterations == maxIterations) {
            nextAccount = list.accounts[address(0)];
            prevId = nextAccount.prev;
            break;
        } else if (nextId == address(0)) {
            nextAccount = list.accounts[address(0)];
            break;
        } else if (nextAccount.value < value) {
            break;
        } else {
            prevId = nextId;
            nextId = nextAccount.next;
            nextAccount = list.accounts[nextId];
        }
    }
}

list.accounts[id] = Account(prevId, nextId, value);
list.accounts[prevId].next = id;
nextAccount.prev = id;
```
<details>
<summary>gas diff</summary>

</details>

src/DoubleLinkedList.sol
```diff
@@ -79,12 +77,11 @@ library DoubleLinkedList {
     /// @param id The address of the account.
     function remove(List storage list, address id) internal {
```
makcandrov, line 78: this function can be optimized:

```solidity
function remove(List storage list, address id) internal {
    Account storage account = list.accounts[id];
    if (id == address(0)) revert AddressIsZero();
    if (account.value == 0) revert AccountDoesNotExist();

    address prevId = account.prev;
    address nextId = account.next;
    list.accounts[prevId].next = nextId;
    list.accounts[nextId].prev = prevId;

    account.prev = address(0);
    account.next = address(0);
    account.value = 0;
}
```
<details>
<summary>gas diff</summary>

</details>


The line `if (id == address(0)) revert AddressIsZero();` can be removed, but the error thrown would no more be accurate in case of null address.

Tristan22400: I will implement but not sure we will merge the PR as mentioned above

test/TestDoubleLinkedList.t.sol
MerlinEgalite, line 1: can you remove changes to the test @Tristan22400 please?

QGarchery (changes requested):
I think we do not want to merge because of [this discussion ](https://github.com/morpho-org/morpho-data-structures/pull/129#issuecomment-1534338934)

MerlinEgalite:
Agree with you @QGarchery. Sorry about that @Tristan22400 I'm closing this PR

PR #130, opened 2023-05-04
Test folder partitioning

makcandrov:
#122 point 3


Tristan22400 (approved):

MathisGD:
I'm not a big fan of complex folder trees. Checking the initial conversation, I have not seen any evidence showing the benefit of such an re-org. So I'm against (but no veto).

makcandrov:
can be closed if everyone agrees

Tristan22400:
> I'm not a big fan of complex folder trees. Checking the initial conversation, I have not seen any evidence showing the benefit of such an re-org. So I'm against (but no veto).

What could be better in my opinion is to separate the test by datastructure. IT could be more readable. If we want to organize these files.

pakim249CAL:
I think just from a practical standpoint, there aren't that many test files to necessitate a reorganization imo. The morpho-aave-v3 repo definitely needed this though because of the sheer number of tests. If this repo grows further, I think we can reorg at that point.

MathisGD:
>What could be better in my opinion is to separate the test by datastructure. IT could be more readable. If we want to organize these files.

Couldn't good file names do the job ?

QGarchery (approved):

Issue #85, opened 2022-11-12
DLL: use address(0) instead of head and tail


MathisGD:
POC here: https://github.com/morpho-dao/morpho-v1/pull/1255

MerlinEgalite:
Closing this as not planned because of this https://github.com/morpho-org/morpho-data-structures/pull/129#issuecomment-1534338934

MathisGD:
I don't think that this is not planned. This idea is objectively good, it's just that we don't have the bandwidth at the moment to adapt the formal verification to it.

PR #141, opened 2024-07-10
DLL loops to address(0)

QGarchery:
This PR:
- completes #85 
- simplifies the loop in insertSorted
- simplifies the munging, notably extracting out MockDll


src/DoubleLinkedList.sol
```diff
@@ -99,47 +96,19 @@ library DoubleLinkedList {
         if (id == address(0)) revert AddressIsZero();
         if (list.accounts[id].value != 0) revert AccountAlreadyInserted();
 
-        uint256 numberOfIterations;
-        address next = list.head; // If not added at the end of the list `id` will be inserted before `next`.
+        address next = list.accounts[address(0)].next; // `id` will be inserted before `next`.
```
Rubilmax, line 99:
```suggestion
        address next = getHead(list); // `id` will be inserted before `next`.
```

Rubilmax: But maybe it adds another JUMP and it complexifies the bytecode only for readability

src/DoubleLinkedList.sol
```diff
@@ -99,47 +96,19 @@ library DoubleLinkedList {
...
 
-        while (numberOfIterations < maxIterations && next != address(0) && list.accounts[next].value >= value) {
+        uint256 numberOfIterations;
+        for (; numberOfIterations < maxIterations; numberOfIterations++) {
+            if (next == address(0) || list.accounts[next].value < value) break;
             next = list.accounts[next].next;
```
Rubilmax, line 104:
```suggestion
            next = getNext(list, next);
```

src/DoubleLinkedList.sol
```diff
@@ -99,47 +96,19 @@ library DoubleLinkedList {
...
-            else {
-                address tail = list.tail;
-                list.accounts[id] = Account({prev: tail, next: address(0), value: value});
-                list.accounts[tail].next = id;
-                list.tail = id;
-            }
-        }
+        if (numberOfIterations == maxIterations) next = address(0);
+
+        address prev = list.accounts[next].prev;
```
Rubilmax, line 109:
```suggestion
        address prev = getPrev(list, next);
```

Rubilmax (approved):
This is so much simpler

Rubilmax (dismissed):

src/DoubleLinkedList.sol
```diff
@@ -99,47 +96,19 @@ library DoubleLinkedList {
...
-            }
-            // Account is not the new head.
-            else {
-                address tail = list.tail;
-                list.accounts[id] = Account({prev: tail, next: address(0), value: value});
-                list.accounts[tail].next = id;
-                list.tail = id;
-            }
-        }
+        if (numberOfIterations == maxIterations) next = address(0);
```
MathisGD, line 107: nb: it relies on the fact that "numberOfIterations++" is always executed after a loop, wether or not the condition for executing the loop is true or not.

When you think about it, it's actually logical (because the condition can depend on the execution of this expression), and it's the way it works in most langages. But it's not that obvious visually.

an other way to do the same thing, that removes this little thing:
```solidity
while (true) {
    if (next == address(0) || list.accounts[next].value < value) break;
    next = getNext(list, next);    
    numberOfIterations++;
    if (numberOfIterations == maxIterations) {
        next = address(0);
        break;
    }
}
```
also I think that it saves a bit of gas (reusing the end of loop condition to avoid the if after).

I hear that the readability is also somewhat impacted negatively

certora/applyHarnessFifo.patch
```diff
@@ -1,146 +1,25 @@
 diff -ruN DoubleLinkedList.sol DoubleLinkedList.sol
---- DoubleLinkedList.sol	2023-05-23 16:01:50.888803463 +0200
-+++ DoubleLinkedList.sol	2023-05-23 16:44:01.777091764 +0200
-@@ -18,6 +18,8 @@
+--- DoubleLinkedList.sol	2024-08-14 10:10:08.974975205 +0200
++++ DoubleLinkedList.sol	2024-08-14 10:14:10.970436843 +0200
```
MathisGD, line 3: Isn't there a way to remove dates in patch files? they don't bring anything and bring a lot of... diff

certora/helpers/MockDllFifo.sol
```diff
@@ -0,0 +1,105 @@
...
+    // VERIFICATION INTERFACE
+
+    DoubleLinkedList.List public dll;
+
+    uint256 public maxIterations;
+
+    uint256 internal dummy_state_variable;
+
+    function dummy_state_modifying_function() public {
+        // to fix a CVL error when only one function is accessible
```
MathisGD, line 18: is it still needed?

src/DoubleLinkedList.sol
MathisGD: actually it doesn't work if `maxIteration` == 0

certora/applyHarnessFifo.patch
QGarchery: There is a builtin way, it's the `--label` flag, but it does not apply to recursive diff. Fixed with df17d59 instead

certora/helpers/MockDllFifo.sol
QGarchery: Nice catch, it's not needed anymore -> removed in ac6cfd2

src/DoubleLinkedList.sol
colin-morpho: also, an [decent] optimizing compiler should handle this kind of optimizations.

QGarchery: Or we could put the `unchecked { ++numberOfIterations; }` back at the end of the loop. What do you think ?

colin-morpho (dismissed):

src/DoubleLinkedList.sol
QGarchery: Yes, you need to check first if you reached mach iterations.
So it would give
```solidity
while (true) { 
  if (numberOfIterations == maxIterations) { next = address(0); break; }
  if (next == address(0) || list.accounts[next].value < value) break; 
  next = getNext(list, next); 
  numberOfIterations++;  
}
```
which starts to be very close to the code of the PR.
Maybe an intermediate solution would be 
```solidity
        for (; numberOfIterations < maxIterations;) {
            if (next == address(0) || list.accounts[next].value < value) break;
            next = getNext(list, next);
            numberOfIterations++
        }

        if (numberOfIterations == maxIterations) next = address(0);
 ```
to make it obvious that `numberOfIterations` is incremented. We could even make it `unchecked`, which would be closer to the code before this PR.

What do you think ?

File: certora/makefile [hunk 7]
```diff
@@ -7,15 +7,15 @@ munged-simple: $(wildcard ../src/*.sol) applyHarnessSimple.patch
 	@patch -p0 -d munged-simple < applyHarnessSimple.patch
 
 record-simple:
-	diff -ruN ../src munged-simple | sed 's+\.\./src/++g' | sed 's+munged-simple/++g' > applyHarnessSimple.patch
+	diff -ruN ../src munged-simple | sed 's+\.\./src/++g' | sed 's+munged-simple/++g' | sed 's,\(\-\-\- [^[:space:]]*\).*,\1,' | sed 's,\(+++ [^[:space:]]*\).*,\1,' > applyHarnessSimple.patch
```
colin-morpho, line 10: Can't we merge this sed filter in a single one? Something along the lines of 
```
sed 's,\(\-\-\- [^[:space:]]*\).*\|\(+++ [^[:space:]]*\).*,\1\2,'
```

QGarchery: I think we can, but I'd say it's harder to read and we don't care much about efficiency for this

QGarchery: I think the following works too: `sed 's,\(\(\-\-\-\|+++\) [^[:space:]]*\).*,\1,'`. Do you like it better than the current one ?

colin-morpho: Yes but is optional do what you thinks is best :) Personally I think your latest proposition is the most readable. In any case, sed filters are not known for their readability 😅!

colin-morpho (approved):

src/DoubleLinkedList.sol
MathisGD: >So it would give

Indeed, that works well (and saves one if). I have mixed feelings, I let you choose

MathisGD (approved):

src/DoubleLinkedList.sol
QGarchery: I'll keep it as is for now. note that we could even do:
```solidity
while (true) { 
  if (numberOfIterations == maxIterations) next = address(0);
  if (next == address(0) || list.accounts[next].value < value) break; 
  next = getNext(list, next); 
  numberOfIterations++;  
}
```

